PR #30, opened 2026-07-26
fix(agent): configure Gemini safety thresholds and steer one regeneration before deflecting

AlaskanTuna:
## Incident

9 `deflection` / `safety` responses on 2026-07-26 13:33–13:48 UTC across 2 guilds and 3 users, every one with `retries = 0` on a clean ~1.2–1.5s round-trip. `/stats` Nerd view's `deflection: 9` was reporting these faithfully.

Chain: Gemini returns a candidate with no content parts → ADK `createLlmResponse` sets `errorCode = candidate.finishReason` (literally `"SAFETY"`) → `roka.ts` short-circuits on `event.errorCode` → `classifyMarker` → `kind: 'safety'` → `retryable: false` → deflects on attempt 0.

**Root cause:** `safetySettings` had never been configured anywhere, so the bot inherited per-model default thresholds — which are not stable across model versions. `gemini-3.1-flash-lite` never exhibited this; `gemini-3.5-flash-lite` does. The existing exponential-backoff retry path was never involved, which is why refining it changed nothing.

## Changes

- **`safetySettings` configured** at `gemini.safetyThreshold` (env `GEMINI_SAFETY_THRESHOLD`), default `OFF`, across the four Gemini-API-supported harm categories. Wired into all three call sites: the agent and **both** memory extractors — a block in an extractor silently dropped memory with no user-visible signal. `HARM_BLOCK_THRESHOLD_UNSPECIFIED` is rejected by validation because it silently reinstates the defaults this change exists to remove.
- **One steered regeneration** on `kind === 'safety'` before the static deflection: an in-character redirect appended to the system prompt via an `AsyncLocalStorage` seam. No backoff sleep (the block is not rate-related), `retryLatencyMs` untouched, consumes an RPM token and degrades to the static line when the floor refuses. `geminiReliability.ts` is deliberately at **zero diff** — `safety` remains `retryable: false, deflect: true`; the one-shot branch lives only in the orchestrator, so there is no blind re-roll.
- **Observability:** `failure_marker` column on `response_events` (guarded additive migration, `tools_used` precedent) plus a per-attempt `Live turn attempt failed` warn. The marker stores only the raw `errorCode`/`finishReason` token, 64-char cap — never `errorMessage`, since a 400 can echo request content. It separates configurable `SAFETY` blocks from server-side `PROHIBITED_CONTENT`/`SPII`/`BLOCKLIST` ones, which is what decides whether thresholds or steering is the applicable lever.
- **Docs:** TRD reliability contract (including the `session_corrupt` row missing since #24) and the README tunable.

## Notes for review

- `src/agent/roka.ts` reports 227 changed lines, but `git diff -w` shows 75 insertions / 16 deletions — the remainder is re-indentation from the `steeringForRequest.run` nesting.
- A `stateDelta`-based steering approach was implemented and rejected: ADK applies `stateDelta` only inside `if (newMessage)` (`runner/runner.js:107-128`) and retries send `newMessage: undefined`, so it would have been silently dropped and the "steered" attempt would have degraded into an unsteered re-roll.
- The per-harm-category rating is not recoverable — `createLlmResponse` drops `candidate.safetyRatings` at the ADK boundary.

## Metrics semantics

| Outcome | `outcome` | `kind` | `retries` | `failure_marker` |
| --- | --- | --- | --- | --- |
| Steered regeneration succeeds | `ok` | `ok` | 1 | `SAFETY` |
| Both attempts blocked | `deflection` | `safety` | 1 | `SAFETY` |
| Regeneration never ran (RPM floor) | `deflection` | `safety` | 0 | `SAFETY` |

## Verification

685/685 vitest passing (0 failures) · `tsc --noEmit` clean · Biome clean (168 files). QA verdict: Approve with comments; 9 findings, all LOW, the two consequential ones fixed pre-merge.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Gemini `safetyThreshold` with environment-based overrides (defaulting to `OFF`).
  * Applied safety thresholds across Gemini generation and memory extraction.
  * Added support for safer outcomes: safety-related failures can trigger guided regeneration when available.
  * Response metrics and stored response events now include a concise `failureMarker` (e.g., safety or rate-limit).

* **Bug Fixes**
  * Improved reliability handling for safety deflections and transient failures, including more robust session recovery and clearer failure attribution.

* **Documentation**
  * Updated Gemini tunables and reliability/failure-marker guidance to reflect the new behavior.


AlaskanTuna:
@coderabbitai review

Issue #39, opened 2026-07-29
fix(agent): recall_user over-fires on messages with no member reference


AlaskanTuna:
## Summary

The Phase 21 live tool-trigger benchmark (issue #28) found that `recall_user` **over-fires** on messages that reference no server member. Proactive triggering itself is sound — recall is a perfect 1.00 — but precision measured 0.667–0.750 across three calibration runs, and two labelled should-not-fire cases fail systematically.

This is the `recall_user` instance of the over-trigger problem issue #19 item 6 anticipates for `search_web`.

## Evidence

Three calibration runs, 12 cases × 3 trials, `gemini-3.5-flash-lite`, 2026-07-29:

| Run | Accuracy | Precision | Recall | Systematic failures |
| --- | --- | --- | --- | --- |
| 1 | 0.8333 | 0.750 | 1.000 | N5 |
| 2 | 0.8056 | 0.720 | 1.000 | N5 |
| 3 | 0.7500 | 0.667 | 1.000 | N1, N5 |

Per-case fired counts (out of 9 across all three runs):

| Case | Label | Message | Fired |
| --- | --- | --- | --- |
| F1–F6 | should fire | (six member-reference variants) | **9/9 each** |
| N1 | should NOT | "ugh it's raining again and I still haven't done laundry" | **7/9** |
| N2 | should NOT | speaker talking only about themselves | 2/9 |
| N3 | should NOT | "I finally finished Frieren last night" | 1/9 |
| N4 | should NOT | pure greeting/banter | 0/9 |
| N5 | should NOT | "Roka, do you ever get nervous talking in front of a crowd?" | **9/9** |
| N6 | should NOT | generic opinion question, no person | 3/9 |

## Root cause — proven, not inferred

A live mutation probe deleted **only** `src/agent/prompts/core.ts:65` (the proactive rule), ran the benchmark once, and restored the line:

| Metric | Rule present | Rule deleted |
| --- | --- | --- |
| Recall | 1.000 | **0.111** |
| Accuracy | 0.750–0.833 | 0.556 |
| N1 fired | 7/9 | **0/3** |
| N5 fired | 9/9 | **0/3** |

Both false positives vanished when that single line was removed, so `core.ts:65` is the cause of the over-firing as well as the correct firing. The rule reads *"Whenever another server member comes up by name"*, and the model is applying it to messages where no member comes up by name at all.

[...]

Issue #19, opened 2026-07-23
enhancement: refine web search — footnote links, dialogue delivery, and provider evaluation


AlaskanTuna:
## Scope

Deep engineering refinement of the web-search feature across **both surfaces** — the `/search` slash command and the implicit `search_web` ADK tool (dialogue invocation). Deferred from the Phase 16 session (2026-07-23) as future work.

### 1. Hyperlink Footnotes
Refine how source links render in the result views for both surfaces — consistent, compact footnote-style citations in the Components V2 container rather than raw/ad-hoc link presentation.

### 2. Dialogue Delivery Optimization
Optimize how Roka weaves web findings into conversational replies when the tool fires mid-dialogue: in-character synthesis quality, source attribution inside the reply, and length behavior under the 50–70-word band (a looked-up answer competing with the word budget is a known tension — see the tool-call turn measured at 102 words in the Phase 15 voice check).

### 3. Web Search Provider Evaluation
Current provider is Tavily (`TAVILY_API_KEY`) and result quality is unsatisfactory. Evaluate replacements (result quality, free-tier limits, latency from the Pi, self-contained ethos — no heavy SDKs). Migration should keep the `search_web` tool contract stable so prompts/handlers don't churn.

## Constraints
- Both surfaces share the executor — refactors must keep slash + implicit behavior consistent.
- Hot-path latency budget applies to dialogue invocation.
- Tool-usage footer (Phase 16, task 113) already indicates `search_web` fired ("searched the wider world") — footnote design should complement, not duplicate it.

## Origin
Requested by the operator during the Phase 16 PM session; intentionally excluded from the Phase 16 PR to keep that batch scoped.

### 4. Invocation Intelligence (Proactive, Uncertainty-Triggered Search)
Roka only searches when the message literally says "search"; on niche factual questions she answers from pretrained knowledge (often wrong) or deflects. Root cause is self-inflicted (session 2026-07-24 diagnosis):

- `search_web` description reads `FALLBACK: ... only when the user explicitly asks about recent news or current events` — forbids proactive use.
- `src/agent/prompts/core.ts` Hard Boundaries line "If asked something you wouldn't know as Roka, deflect naturally in character" actively routes answerable factual questions away from search.

Fix (same pattern as the `recall_user` proactive-trigger fix, PR #21 era):
[...]

PR #40, opened 2026-07-29
test(agent): benchmark proactive recall_user firing against the live model

### LINKED ISSUE #28: test(harness): smoke-test proactive recall_user triggering via dev harness
## Context

The proactive `recall_user` triggering shipped in PR #21 (description enumerating trigger cases, core-prompt rule, mention→display-name rewriting) was verified live but never benchmarked systematically. The operator requested a harness smoke test ("can be smoke tested in our dev harness later using Codex") — deferred, never scheduled.

## Task

Add a harness eval that runs a dialogue set against the real prompt + tool descriptions and asserts whether `recall_user` fires:

- **Should fire:** direct questions ("what do you know about X"), stories/claims mentioning a member ("X promised me to attend this seminar last week"), @mention variants (post-rewrite they arrive as `@displayName`), members with zero stored notes (tool should still fire).
- **Should NOT fire:** messages with no member reference, the speaker talking only about themselves, generic names that match no known member.

## Notes

- Same benchmark pattern as the existing harness evals; Codex worker can implement (standard invocation contract; pre-decline optional interactive features in the brief).
- Overlaps with issue #19 item 6, which needs the same should-fire/shouldn't-fire benchmark shape for `search_web` — consider building the dialogue-set fixture generically so both tools share the rig.
- Costs real Gemini calls per run — keep the dialogue set small and exclude from the default `npm test` gate (align with wherever the perf evals land per the memoryShadow isolation issue).

AlaskanTuna:
Closes #28.

Adds the repo's first deliberately-networked test: a benchmark that runs a labelled dialogue set against the **real** model, prompt, and tool declarations and asserts whether `recall_user` fires. Every existing harness eval stubs the model (`__setTestRunTurnFactory`, `vi.mock('@google/adk')`) — that seam is exactly what this cannot use, because the subject under test *is* whether the real model decides to call the tool.

## What it answers

**Proactive `recall_user` triggering works.** Recall is **1.00** across three calibration runs — all six should-fire cases fired 3/3, every run, 54 for 54. PR #21's rule does what it claims.

| Run | Accuracy | Precision | Recall | Systematic failures |
| --- | --- | --- | --- | --- |
| 1 | 0.8333 | 0.750 | 1.000 | N5 |
| 2 | 0.8056 | 0.720 | 1.000 | N5 |
| 3 | 0.7500 | 0.667 | 1.000 | N1, N5 |

## The gate ships red, on purpose

The benchmark found the *inverse* defect: `recall_user` over-fires on messages referencing no member — N1 (ambient chatter, 7/9) and N5 (a message addressing Roka herself, 9/9). Filed as **#39**. Operator decision was to ship the gate reporting the real defect rather than relabel cases to make it pass; the fix follows as separate work.

`npm test` and `npm run test:perf` are unaffected and green. Only the opt-in, paid `npm run test:live` is red, and `docs/README.md` says so plainly with the tracking issue.

## Root cause proven by mutation, not inferred

Deleting **only** `src/agent/prompts/core.ts:65` and re-running:

| Metric | Rule present | Rule deleted |
| --- | --- | --- |
| Recall | 1.000 | **0.111** |
| Accuracy | 0.750–0.833 | 0.556 |
| N1 fired | 7/9 | 0/3 |
| N5 fired | 9/9 | 0/3 |

Both false positives vanished with the rule, so `core.ts:65` causes the over-firing *and* the correct firing. This also killed the competing explanation for N5 — `tools/index.ts` separately licenses a no-arg self-recall, which would have made N5 a mislabelled fixture. That description was untouched by the probe and N5 still went to 0/3, so the model was reading "Roka" as a member name. The label is correct. `src/` was restored and verified byte-clean.

## Design

- **Third gate.** `npm run test:live` + `vitest.live.config.ts`, excluded from the default gate by glob, reusing PR #38's `test:perf` pattern. Three gates, three non-overlapping meanings — a network hiccup can never read as a correctness or latency regression.
- **Tool-generic from the start.** Issue #19 item 6 needs the identical shape for `search_web`; it can drop in a `search-web.jsonl` with no code change. The rig also unsets `TAVILY_API_KEY` for the run so that stays true.
- **Statistical verdict, not per-case.** Accuracy floor plus zero systematically-wrong cases. Per-case majority voting flakes ~22% per case at `temperature: 0.9`; with 12 cases something would fail most runs. `MIN_ACCURACY` is **measured, not invented** — pinned to 0.75 from the three runs above.
- **The offline mutation pin.** `toolTrigger.scoring.test.ts` feeds the pure scorer never-fires / always-fires / marginal / one-bad-case observations and asserts the verdict flips — free, in the default gate, on every run. A gate nobody has seen fail is not a gate.
- **Abort, never score, a corrupted turn.** A rate-limited or deflected turn returns the fallback with empty `toolsUsed`, which would score as a genuine no-fire and bias accuracy downward invisibly. The rig throws on `metrics.outcome !== 'ok'`, naming case, trial, outcome and kind.

## Review round found two things that would have cost money

- **The guard was blind to `terminal`.** An expired or revoked `GRAPHIFY_GEMINI_API_KEY` classifies as `terminal` → `deflection`, which passed through as a real observation. All 36 trials would have returned accuracy exactly 0.500 with all six should-fire cases flagged — byte-identical to a genuinely broken proactive rule, manufacturing a false accusation against shipped behaviour after burning the quota.
- **The fixture collided with its own seeded world.** Kaede was seeded `favorite_anime: "Frieren"` while case N3 read "I finally finished Frieren last night". Not passive — the retriever FTS-matches claim values and promoted Kaede from third to second on that case alone. Kaede's claim is now `Mushishi`.

Also fixed: the verdict rule was duplicated across the live and scoring tests, so calibrating the threshold would have silently desynced the mutation pin. Now one exported `MIN_ACCURACY` + `meetsLiveVerdict`, with a coupling test deriving its boundary from the imported constant.

## Verification

`npm test` **715 / 57 files** · `npm run test:perf` **2** · `tsc --noEmit` clean · Biome clean (174) · `npm run format:check` clean · **no file under `src/` modified** across the whole phase.

Cost of a live run, measured: ~60 Gemini calls and ~8.5 minutes, paced at 10 RPM against the 15 RPM cap. Runs on `GRAPHIFY_GEMINI_API_KEY`, never the bot's.

## Note for reviewers

`npm run test:live` is **not** part of full verification, which remains `npm test && npm run test:perf`. It is opt-in, costs real money, and currently fails by design per #39.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in live-model test command for evaluating tool-trigger behavior against realistic dialogue scenarios.
  - Added fixture-based scoring for accuracy, precision, recall, and systematic failures.
  - Added `recall_user` evaluation scenarios covering expected tool use and non-use.

- **Documentation**
  - Expanded test instructions, live-test requirements, benchmarks, and current results.

- **Tests**
  - Added validation for fixture integrity, scoring decisions, and isolated live-test environments.


AlaskanTuna:
Addressed in `f8c1f08`. Three of four applied, one declined with reasoning.

**Applied**

- **`toolTriggerScoring.ts` — `asClaim` subjectId validation (Major).** Correct and the most valuable of the four. Threaded `memberIds` into `asClaim` the same way `asHistoryLine` already receives it. Mutation-verified: reverting the check drops the suite to 16/17, failing exactly the new `'a claim subjectId outside members'` scenario.
- **`toolTriggerScoring.ts` — fixture line-number drift (Minor).** Correct. Now enumerates before filtering so the reported number is the true file line. New test writes header / blank / malformed-case and asserts line **3**; the pre-fix code reported line 2, which is what proves the fix.
- **`toolTrigger.live.test.ts` — empty `GEMINI_API_KEY` (Minor).** Applied. Defence-in-depth rather than a live hole, since `tests/harness/env.ts` already throws on a missing `GRAPHIFY_GEMINI_API_KEY` before assignment — but the preflight exists to fail before money is spent, so the stricter assertion belongs there. The sibling `GOOGLE_GENAI_API_KEY` assertion already used `toBeTruthy()`.

**Declined — `toolTrigger.ts:42`, clear SQLite session history on teardown**

The premise doesn't hold here. It rests on repeated runs accumulating `session_history` rows in `data/rokabot.db`, but the harness never writes that file: `tests/harness/env.ts:6` pins `ROKABOT_DB_PATH = ':memory:'`, and the live test's preflight asserts that exact value before anything runs, so the database dies with the process.

Independently, accumulation can't occur within a run either — channel ids are `live-${testCase.id}-${trial}`, unique per trial and seeded exactly once each (`toolTrigger.ts:85-86`). Adding `clearHistory` would be a no-op against an in-memory DB that is already discarded.

Gates after the fixes: `npm test` **717 / 57 files**, `npm run test:perf` 2, `tsc --noEmit` clean, Biome clean (174), `format:check` clean. No file under `src/` touched.

Issue #35, opened 2026-07-29
memoryShadow.eval.test.ts flakes on vitest's default 5000ms testTimeout


AlaskanTuna:
## Summary

`tests/harness/__tests__/memoryShadow.eval.test.ts` sits almost exactly on vitest's default 5000ms `testTimeout` and fails intermittently when the full suite runs under CPU load.

## Evidence

- In isolation the test completes in **2183 / 2233 / 2241 ms** across runs.
- In a clean full-suite run it passed at **exactly 5000 ms**.
- Under additional load it tips over:

```
 Test Files  1 failed | 56 passed (57)
      Tests  1 failed | 684 passed (685)
 FAIL  tests/harness/__tests__/memoryShadow.eval.test.ts
Error: Test timed out in 5000ms.
```

No `testTimeout` is configured in `vitest.config.ts`, so the suite inherits vitest's 5000ms default.

## Why it is worth fixing

This already caused a misdiagnosis. During the issue #32 work the failure was initially attributed to the `rtk npm test` wrapper "mis-parsing a large log" and dismissed as tooling noise. It was a real vitest failure the whole time. A flake that presents as tooling noise invites the next contributor to wave off a genuinely red suite — or worse, to "fix" a test that was passing.

CI is unaffected: `.github/workflows/deploy.yml` does not run the suite. This bites local runs and agent workflows only.

## Proposed fix

Pin an explicit `testTimeout` in `vitest.config.ts` with enough headroom above the observed ~2.2s runtime that ordinary load cannot cross it. Prefer a suite-wide default over a per-test override so other near-boundary tests benefit too.

## Acceptance criteria

- [ ] `testTimeout` is set explicitly rather than inherited from vitest's default.
- [ ] `memoryShadow.eval.test.ts` passes consistently under a full-suite run with concurrent CPU load.
- [ ] The chosen value is justified against the observed runtime, not picked arbitrarily.

Issue #27, opened 2026-07-23
test(harness): isolate memoryShadow perf eval from parallel-load p95 flake


AlaskanTuna:
## Context

`tests/harness/__tests__/memoryShadow.eval.test.ts` asserts a wall-clock performance bound (`expect(report.p95LatencyMs).toBeLessThan(10)`). Under vitest's parallel worker load in a full `npm test` run, CPU contention occasionally pushes p95 past 10ms and fails the suite. Known flake signature: it is the **sole** failure in a full run, and a standalone rerun of the file is green. Documented across recent sessions but never fixed.

## Task

Stop a timing assertion from flaking the whole gate. Candidate approaches (pick one at implementation time):

- Mark the eval as sequential/isolated so it never shares CPU with other workers (e.g. vitest `poolMatchGlobs` / `sequence` config, or moving perf evals to a separate `npm run test:perf` script excluded from the default gate).
- Alternatively, loosen the assertion under detected parallel load — less preferred, since it weakens the exit criterion the number exists to protect.

## Acceptance

- Full `npm test` no longer fails on this assertion under load (verify with several consecutive full runs).
- The 10ms exit criterion itself stays enforced somewhere that runs in CI — isolation, not deletion.

PR #46, opened 2026-07-29
chore(config): gate the Pi deploy on the test suite and run it on pull requests

### LINKED ISSUE #45: ci(deploy): main auto-deploys to the Pi with no test gate
## Summary

`.github/workflows/deploy.yml` triggers on **push to `main`** and runs `docker compose up -d --build` on the Pi's self-hosted runner. It has no test step. A merge to `main` therefore builds and ships to production without `npm test` ever running in CI.

Combined with an already-recorded fact — the repo has **no CI test job and no pre-commit test hook** (noted when `npm run test:perf` was documented in PR #38) — the practical position is that nothing automated ever runs the test suite, yet something automated always deploys.

## Why this matters more than it looks

The suite is the only thing standing between a bad merge and production, and it currently runs solely at the discretion of whoever is working. That is fine while every change goes through a careful review, and silently fatal the first time one doesn't. The failure mode is not "tests fail and we notice" — it is "tests were never run and the bot is already serving the regression."

There is a live precedent for the near-miss: on 2026-07-28 a genuinely failing test was waved off as tooling noise (see #35). Under this pipeline, a merge in that state would have deployed.

## Current behaviour, confirmed

Recent `deploy.yml` runs, all `success`, each immediately following a squash-merge:

| Commit | Deployed |
| --- | --- |
| `5a0d18d` | 2026-07-29T16:51Z |
| `0dcda5c` | 2026-07-29T16:22Z |
| `7daa222` | 2026-07-29T10:55Z |

The workflow does run a health check after `compose up` and notifies Discord on success or failure, so a container that fails to start is caught. What is not caught is a container that starts perfectly and behaves wrongly.

## Suggested fix

Add a test job to the workflow, gating the deploy step on it:

- `npm ci`, `npx tsc --noEmit`, `biome check .`, `npm test` — all offline, all free, no API key required.
- Deploy only `if: success()`.
- Deliberately **exclude** `npm run test:live` (needs `GRAPHIFY_GEMINI_API_KEY`, spends real Gemini quota, and currently fails by design per #39). `npm run test:perf` is wall-clock sensitive and would be flaky on a shared Pi runner competing with the build — either skip it or run it on `fileParallelism: false` with the existing 30s ceiling and accept some noise.
- Consider running the test job on a GitHub-hosted runner rather than the Pi, so a test run cannot contend for CPU with the production container it is about to replace. The Pi has 8GB and is running the live bot; `npm ci` plus a full suite alongside it is exactly the CPU contention that caused #27 and #35.

## Acceptance

- [ ] A merge to `main` whose test suite fails does not reach the Pi.
- [ ] The deploy step is explicitly gated on the test job's success.
- [ ] Live and perf gates are handled deliberately (excluded or isolated), not swept in by accident.

AlaskanTuna:
Closes #45.

`deploy.yml` fired on push to `main` and ran `docker compose up -d --build` on the Pi with **no test step**. The repo has no separate CI test job and no pre-commit test hook either, so nothing automated ever ran the suite while something automated always deployed. A merge with a failing suite shipped straight to production.

This was not hypothetical: on 2026-07-28 a genuinely failing test was waved off as tooling noise (#35). Under this pipeline that state would have deployed.

## What changed

- New `test` job: `npm ci` → `tsc --noEmit` → `biome check .` → `npm test` → `npm run test:perf`.
- `deploy` now carries `needs: test`.
- Tests also run on `pull_request`, so a failure surfaces at PR time rather than at merge time.

Every pre-existing deploy step is untouched — the diff is **39 insertions, 0 deletions**. The pull, the build, the health-check retry loop and both Discord notifications are byte-identical; the only change inside the `deploy` job is the two added keys.

## Two deliberate choices

**Tests run on `ubuntu-latest`, not the Pi.** Running the suite on the self-hosted runner would put it in CPU contention with the live production container it is about to replace — and that exact contention is what produced #27 and #35, where a wall-clock-sensitive eval failed under load. Isolating the test job removes the failure mode rather than inheriting it. The repo is public, so hosted minutes are free.

**`deploy` cannot run from a pull request.** `if: github.event_name == 'push'` is evaluated before the job starts, so a PR run — including from a fork — can never reach the self-hosted runner or the Pi. On a public repo that guard is load-bearing, not cosmetic. `needs: test` then independently blocks a `push` run whose suite fails.

`npm run test:live` is deliberately **excluded** and must stay so: it needs `GRAPHIFY_GEMINI_API_KEY`, spends real Gemini quota, and currently fails by design per #39.

## On the perf gate

`npm run test:perf` is included and therefore blocking. It asserts `p95 < 10ms` and is wall-clock sensitive, so this is a real trade: a flaky hosted runner could block a deploy. Including it is the more consistent choice — this repo's own recorded lesson is that a gate nothing runs is a dead gate, and excluding it would preserve exactly the status quo this issue exists to fix. It measures ~1.4 ms locally and only breached the threshold under 4× CPU oversubscription, so a dedicated 2-vCPU runner has ample headroom. Worth watching; easy to loosen if it proves noisy.

Note the related flake is **not** in the blocking main gate: `memoryShadow.eval.test.ts` is excluded from `npm test` and runs only under `test:perf`, isolated with `fileParallelism: false` and a 30s ceiling (PR #38). Verified — `vitest list` on the default config returns zero matches for it.

`biome` is invoked as `node_modules/.bin/biome`, not `npx biome`, per the recorded 2026-07-24 workaround: `npx` resolves the bare name `biome` rather than the scoped `@biomejs/biome` that is actually installed, and fails with `could not determine executable to run`. That reproduces on a bare invocation with no shell wrapper, so it is expected to affect a clean CI checkout too.

## Verification

Workflow YAML parses; job graph confirmed as `test` (no deps) → `deploy` (`needs: test`, `if: github.event_name == 'push'`); triggers are `push` and `pull_request`, both carrying the same `paths-ignore` list so doc-only changes burn no CI.

The exact commands the job will run, executed locally on a clean tree: `tsc --noEmit` clean · `biome check .` clean (174 files) · `npm test` **722 / 57 files** · `npm run test:perf` **2** · `format:check` clean (the new YAML is Prettier-formatted, since `.prettierignore` does not exclude `.github/`).

This PR validates itself: merging it triggers a `push` run of the very workflow it adds, so the test job gates its own deploy.

## Note on the commit type

Committed as `chore(config)` rather than `ci`. `commitlint.config.js` restricts `type-enum` to `feat|fix|refactor|docs|test|chore|style|perf`, and `AGENTS.md`'s Git Commit Convention lists the same set — neither includes `ci`. The husky `commit-msg` hook correctly rejected it. Worth considering whether `ci` belongs in that enum now that the repo has a real pipeline, but that's a convention change and not this PR's business.

## Summary by CodeRabbit

* **Tests**
  * Automated tests now run for pull requests targeting the main branch.

* **Bug Fixes**
  * Deployments are restricted to push events and occur only after tests complete.
  * Pull requests no longer trigger deployments.


AlaskanTuna:
Addressed in `150695a`. Both the security finding and a CI failure the first run exposed.

**CodeRabbit's Major finding — applied, and it was a real hole.** The `test` job runs pull-request-authored code (including from forks, on a public repo) through `npm ci`, whose postinstall scripts would have inherited a persisted `GITHUB_TOKEN`. That is a supply-chain exposure I introduced while closing a CI gap, not a lint nit. Now `permissions: contents: read` at workflow level (covering both jobs — `deploy` needs nothing more, since it uses its own local git rather than `actions/checkout`, and `secrets.DISCORD_WEBHOOK_URL` is unaffected because `permissions:` governs only the `GITHUB_TOKEN`), plus `persist-credentials: false` on the test job's checkout.

**Separately, the first run failed for a reason worth recording.** `Missing required environment variable: DISCORD_TOKEN` from `src/config.ts:11`. The suite has never been runnable without a `.env` — every local run has one, and `src/config.ts:3` imports `dotenv/config`, so this was invisible until a clean checkout ran it. Three vars are `requiredEnv`: `DISCORD_TOKEN`, `DISCORD_CLIENT_ID`, `GEMINI_API_KEY`. All three now get `ci-dummy-` placeholders at job level — deliberately shaped so a grep for a real Discord token or `AIza` key finds nothing. No repository secrets, no change to `.env.example`, and `requiredEnv` untouched: the tests need config to *resolve*, not to be real, and nothing in `npm test` makes a live call.

Verified by genuine reproduction rather than a same-shell run: a detached `git worktree` with no `.env`, `npm ci`, then `env -i` with only `PATH`/`HOME` and the three placeholders — **722 tests / 57 files** pass, `test:perf` 2 pass.

**The gate is verified working, not merely asserted.** On the failing run `test` reported `fail` and `deploy` reported `skipping`, so both `needs: test` and `if: github.event_name == 'push'` demonstrably held. Current run: `test` pass in 38s, `deploy` skipped.

Issue #60, opened 2026-07-31
fix(agent): the cumulative backoff-cap bail-out is a fourth unmodelled retry bound


AlaskanTuna:
Found by QA while reviewing #58's fix (2026-07-30). Not blocking that PR — the shipped `liveMaxRetries` is `2`, well inside the safe range — but it is the **fourth** appearance of the same defect class (#41 -> #43 -> #58 -> this).

## The bound

`src/agent/roka.ts:325` carries a cumulative backoff-cap bail-out: the retry loop stops once accumulated `retryLatencyMs` reaches `retryBackoffCapMs`.

At the shipped `retryBackoffBaseMs: 1000` / `retryBackoffCapMs: 12000`, the worst-case backoff schedule is 1000, 2000, 4000, 8000 — cumulative 15000, which crosses 12000 on the fourth retry. So **`liveMaxRetries >= 5` can never produce more than 5 attempts at *any* `turnDeadlineMs`.**

QA verified this by simulating the real loop: at `R = 0..4` it stops on `retryLimit`; at `R = 5` and `R = 6` it stops on `backoffCapBailout`.

## Why the new guard does not catch it

#58 added a startup guard asserting the *deadline* bound:

```
requiredMs = (liveMaxRetries + 1) * timeout + Σ min(retryBackoffBaseMs * 2^k, retryBackoffCapMs)
```

warned on at `turnDeadlineMs < requiredMs`. That models the turn-deadline tail only. At `liveMaxRetries: 5`, `requiredMs` is 147000 — so an operator who sets `turnDeadlineMs: 150000` gets **no warning at all**, while the loop still silently caps at 5 attempts because of the backoff-cap bail-out.

That is exactly the shape of the original defect: a retry-budget relationship asserted nowhere enforceable, silently violated by a later tuning change.

## Suggested work

Extend the startup guard to model *both* tails, or add a second warning for the backoff-cap bound: warn when `Σ_{k=0}^{liveMaxRetries-1} min(retryBackoffBaseMs * 2^k, retryBackoffCapMs) >= retryBackoffCapMs`, i.e. when the configured retry count cannot be reached before the cumulative cap fires.

[...]

Issue #59, opened 2026-07-31
chore(agent): surface the HTTP status code in the failure marker so incidents are greppable


AlaskanTuna:
Deferred from #58 (Gate 1 Q4(2), 2026-07-30) rather than dropped.

## Problem

`markerFrom` (`src/agent/roka.ts:154-158`) logs `errorName`, so a Gemini failure surfaces as `marker: "ApiError"`. The HTTP status lives only in `errorMessage`.

During the 2026-07-30 production incident this meant grepping the Pi's logs for `504` found nothing — the code was there, but not in any field the log lines are indexed on. Classifying the incident took materially longer than it needed to.

## Why it was deferred rather than done in #58

Three reasons, in weight order:

1. **Wrong file for that PR's scope fence.** `markerFrom` is in `roka.ts`, not `geminiReliability.ts`. Adopting it would have collided with #52's file scope and broken both the parallel wave and the PR split.
2. **It brushes a recorded decision.** `docs/decisions.md:64` says `failure_marker` persists only the raw `errorCode`/`finishReason` token and never `errorMessage`, because a Gemini 400 can echo request content back. A digit-anchored 3-character status extracted from the message is arguably *not* "errorMessage" — but that is a deliberate ruling, not an assumption to make in passing.
3. The parent issue worded it "Consider", and the value is log greppability rather than behaviour.

## Work

- Decide whether a digit-anchored status extraction is compatible with `decisions.md:64`'s privacy constraint. If it is not, a **log-only** variant (status in the log line, not in the persisted `failure_marker`) is the cheaper alternative and is probably the right scope.
- If adopted, reuse the existing digit-anchoring convention from `geminiReliability.ts` (`/(?<!\d)(?:…)(?!\d)/`) so a status cannot be matched inside a larger number.
- Pin it with a test whose fixture does not also carry a network token — see the trap documented in #58's PR.

Low priority. Observability, not correctness.

Issue #41, opened 2026-07-29
fix(config): turnDeadlineMs of 60s makes retries unreachable at a 45s gemini.timeout


AlaskanTuna:
## Summary

`gemini.timeout: 45000` and `gemini.turnDeadlineMs: 60000` are mutually inconsistent: after a first attempt times out, no retry can ever be admitted, so `gemini.maxRetries: 3` is unreachable for timeout-class failures — the most common failure mode during upstream degradation.

Observed in production during a 9.6-minute degradation window on 2026-07-29 22:56–23:05 +08.

## Evidence

Five fallback responses were served. Four were `kind: network` with `llmMs ≈ 44,900` (i.e. hitting the 45s request timeout) and **`retries: 0`** in every case:

```
{"outcome":"fallback","kind":"network","failureMarker":"ApiError",
 "llmMs":44932,"retries":0,"retryLatencyMs":0,"e2eMs":46882}
```

The admission gate logged its reasoning each time:

```
{"attempt":0,"elapsedMs":44874.6,"delayMs":747.3,"deadlineMs":60000,
 "requestTimeoutMs":45000,"kind":"network",
 "msg":"Turn deadline would be exceeded by planned retry backoff"}
```

No crash, no OOM, no restart (`restarts=0`, uptime unbroken). This is not an availability incident in the bot.

## Mechanism

The pre-spend gate added in #36 admits a retry only when `remainingMs >= delayMs + requestTimeoutMs`. After a full timeout:

- `elapsedMs ≈ 44,900`
- `remainingMs = 60,000 − 44,900 = 15,100`
- required `= ~750 + 45,000 = 45,750`
- `15,100 < 45,750` → refuse, always

For **one** retry to be admissible after a full timeout, the budget must satisfy roughly `turnDeadlineMs >= 2 × timeout + backoff`, i.e. **≥ ~91s** at the current timeout. At 60s it is unreachable by a wide margin.

**The gate is behaving correctly** — this is a configuration-consistency defect, not a regression in #36. Before #36 the same failure retried and produced the 96.8s / 137.8s turns that motivated the deadline in the first place. The current behaviour (fail at ~46s) is better for channel throughput; the problem is that it forfeits the retry entirely rather than making it fit.

## Why the existing guard did not catch it

`src/config.ts:221` warns only when `turnDeadlineMs <= timeout`. Here `60000 > 45000`, so no warning fires — but the *effective* requirement for a usable retry path is far stricter than "greater than".

[...]

PR #61, opened 2026-07-31
fix(config): budget the turn deadline for every promised retry attempt

### LINKED ISSUE #58: fix(agent): the third live retry attempt is structurally unreachable, so liveMaxRetries: 2 behaves as 1
## Summary

`gemini.liveMaxRetries: 2` promises 3 attempts. Under degradation — precisely when retries matter — **the third attempt is structurally unreachable**. This is the same class of defect as #41, moved one slot along: #41's fix made the *first* retry reachable, but the identical arithmetic still blocks the *second*.

Found while diagnosing a live fallback message on 2026-07-30 (guild 698854883872210944, 14:03:59–14:04:41 UTC).

## The arithmetic

`config.yml`: `timeout: 20000`, `turnDeadlineMs: 60000`, `liveMaxRetries: 2`, `retryBackoffBaseMs: 1000`, `retryBackoffCapMs: 12000`.

The deadline gate (`src/agent/roka.ts:336-339`) refuses a retry when `remainingMs < delayMs + requestTimeoutMs`.

When each attempt consumes its full timeout:

| Point | Elapsed | Remaining | Backoff range | Need | Result |
| --- | --- | --- | --- | --- | --- |
| before attempt 2 | 20000 | 40000 | 500–1000 | 20500–21000 | reachable |
| before attempt 3 | 40000 | **20000** | **1000–2000** | **21000–22000** | **always refused** |

**The backoff has a floor.** `src/agent/geminiReliability.ts:104` returns `cappedDelay * (0.5 + randomValue * 0.5)` — half-jitter, not full jitter. The delay before attempt 3 is therefore in `[1000, 2000]` and can never approach zero. So `remaining (20000) < delay (>=1000) + timeout (20000)` holds **unconditionally**, not merely usually.

Total wall-clock needed for 3 attempts: best case `3*20000 + 500 + 1000 = 61500ms`, worst `63000ms` — **even the best case overruns the 60000ms deadline by 1.5s.**

## Observed instance

```
{"level":40,"attempt":0,"kind":"network","marker":"ApiError","msg":"Live turn attempt failed"}
{"level":40,"attempt":1,"kind":"network","marker":"ApiError","msg":"Live turn attempt failed"}
{"level":40,"attempt":1,"elapsedMs":40333.7,"delayMs":1028.1,"deadlineMs":60000,"requestTimeoutMs":20000,
 "kind":"network","msg":"Turn deadline would be exceeded by planned retry backoff"}
{"level":30,"e2eMs":41877,"generateMs":40838,"retries":1,"outcome":"fallback","kind":"network",
 "failureMarker":"ApiError","msg":"Response completed"}
```

Both attempts returned `{"error":{"code":504,"message":"Deadline expired before operation could complete.","status":"DEADLINE_EXCEEDED"}}`. Not a 429, not quota — it was the only request in 5 hours. Container healthy throughout (up 4h, 64MB/512MB, 0 restarts).

## Fix options

- `turnDeadlineMs >= 63000` — covers the worst-case backoff, keeps the 20s timeout that #43 calibrated against measured p95.
- `timeout <= 19000` — worst case becomes exactly 60000ms. Tighter, but re-opens the question #43 settled (20s was chosen as ~2x measured p95 of 9801ms with headroom for the tool chain).

Either way, the **general** fix is that this relationship should be asserted rather than rediscovered: `config.ts` already has a startup guard for `turnDeadlineMs <= 2 * timeout` (added in #44). That guard encodes the *first* retry's reachability. It should encode the reachability of **all `liveMaxRetries` attempts**, i.e. roughly `turnDeadlineMs >= (liveMaxRetries + 1) * timeout + sum(worst-case backoffs)`. Otherwise raising `liveMaxRetries` again will silently produce the same dead configuration a third time.

## Secondary findings

1. **`504` is not in `TRANSIENT_STATUS_PATTERN`.** `src/agent/geminiReliability.ts:34` is `/(?<!\d)(?:429|500|503)(?!\d)/`. The 504 was classified retryable only because `DEADLINE_EXCEEDED` matched `NETWORK_PATTERN` at `:31` first, which is checked at `:64` before the status pattern at `:66`. Correct outcome, but by string coincidence rather than by status code — a 504 whose message worded differently would be treated as terminal.
2. **`marker` logs `errorName`, hiding the status code.** The marker is `"ApiError"`, so grepping the logs for `504` finds nothing; the code is only in `errorMessage`. This made the incident harder to classify than it needed to be.

## Acceptance

- [ ] All `liveMaxRetries + 1` attempts are reachable under full-timeout degradation, demonstrated arithmetically.
- [ ] The startup guard asserts the general relationship, not just the 2-attempt case, so a future `liveMaxRetries` bump cannot silently recreate this.
- [ ] Decide whether `504` should join `TRANSIENT_STATUS_PATTERN` explicitly rather than relying on the message text.
- [ ] Consider surfacing the HTTP status in `marker` so incidents are greppable by status code.

AlaskanTuna:
Closes #58.

## The defect

`gemini.liveMaxRetries: 2` promises 3 attempts. Under degradation — precisely when retries matter — **the third was structurally unreachable**. It caused a live fallback on 2026-07-30 (guild `698854883872210944`, two consecutive HTTP 504 `DEADLINE_EXCEEDED`).

The deadline gate (`src/agent/roka.ts:336-339`) refuses a retry when `remainingMs < delayMs + requestTimeoutMs`. At `turnDeadlineMs: 60000` / `timeout: 20000`:

| Point | Elapsed | Remaining | Backoff | Need | Result |
| --- | --- | --- | --- | --- | --- |
| before attempt 2 | 20000 | 40000 | 500–1000 | 20500–21000 | reachable |
| before attempt 3 | 40000 | **20000** | **1000–2000** | **21000–22000** | **always refused** |

Unconditionally, not merely usually — `computeBackoff` (`geminiReliability.ts:104`) is **half**-jitter (`cappedDelay * (0.5 + random*0.5)`), so the delay has a floor of 1000 and never approaches zero.

This is the **third appearance of the class** (#41 → #43 → #58).

## Two defects in the existing guard, not one

`src/config.ts:298` was `turnDeadlineMs <= 2 * timeout` → `logger.warn`. It was **mis-shaped** (encoded 2 attempts; ignored backoff entirely) *and* **silent** — `60000 <= 40000` is false, so it never fired on the broken shipped config.

## The fix

```
requiredMs = (liveMaxRetries + 1) * timeout
           + Σ_{k=0}^{liveMaxRetries-1} min(retryBackoffBaseMs * 2^k, retryBackoffCapMs)
```

= **63000** at shipped values (max-jitter form). Warned on at `turnDeadlineMs < requiredMs`. `config.yml` `turnDeadlineMs` 60000 → **63000**; `timeout` stays 20000, preserving #43's measured p95 calibration (9801ms, n=78).

QA confirmed `requiredMs` is the **exact inverse** of the admission rule at `roka.ts:339`, matching its strictness at the boundary — 63000 is tight, not merely sufficient.

**It stays a `warn`, not a `throw`.** A throw at startup on a Pi under `restart: unless-stopped` is a crash-loop and a total outage; a cross-key mismatch degrades behaviour rather than breaking it, so the guard's failure mode would be worse than the defect's. Enforcement is fatal at **build** time instead: `src/__tests__/config.test.ts:180` asserts `expect(warn).not.toHaveBeenCalled()` against the **real shipped `config.yml`**.

**That assertion, not the number, is this PR's deliverable** — it makes a fourth recurrence a red CI run rather than a production incident. The guard and the `config.yml` value therefore had to land in the same commit.

## Secondary: `504` joins `TRANSIENT_STATUS_PATTERN`

One line: `429|500|503` → `429|500|503|504`.

**This does not change the incident's outcome.** Those 504s already retried, because their message carried `DEADLINE_EXCEEDED` and `NETWORK_PATTERN` (`:31`) is checked at `:64`, *before* the status patterns at `:65-66`. Both `network` and `transient_http` yield `retryable: true, deflect: false`, so only the reported `kind`/metric label moved.

It is load-bearing **only** for a 504 whose message lacks a network token — which today falls through to `result('terminal')` at `:93`: `retryable: false` **and `action: 'destroy'`**, dropping conversation context on a transient gateway timeout.

502/408 deliberately excluded — no observed instance, and this repo's convention is measured-not-invented.

## Test changes

`roka.test.ts`'s shipped-values test was **pinning the defect** (expected 1 retry / 2 attempts). Renamed to `'reaches all liveMaxRetries + 1 attempts within the shipped turn deadline'`, assertions re-derived from `config.gemini.liveMaxRetries` instead of literals, and the stubbed `computeBackoff: () => 1_000` replaced with the real production function forced to worst-case jitter.

**Latent test defect found in passing:** its `sleep` stub never advanced the fake clock, so the deadline gate was never actually exercised under elapsed backoff. With it advancing, the worst case lands exactly on the boundary — elapsed 41000, remaining 22000, need 22000, gate is strict `<` → admitted. That is what pins 63000 rather than 62000.

## Mutation pinning (2026-07-29 convention)

Every probe produced its predicted failure, then was restored:

- Regex reverted to `(?:429|500|503)` → both new 504 cases fail as `terminal`.
- `turnDeadlineMs` reverted to `60000` → the new roka test fails (1 retry, not 2 — reproducing #58) **and** four `config.test.ts` assertions.
- `50000 → warns` discriminates the **attempt-count** fix (silent under the old `2 * timeout` formula).
- `61000 → warns` discriminates the **backoff term** (silent if it were dropped).

All 504 fixtures were grepped clean for `DEADLINE_EXCEEDED|timeout|abort|fetch failed|ECONNRESET|ETIMEDOUT|EAI_AGAIN` — a fixture carrying any of those short-circuits one tier early and passes for the wrong reason. QA re-pinned all three independently against a scratch copy of the classifier.

## Verification

`npm test` **753 passed / 57 files** (747 baseline + 3 + 3; the rename is net zero) · `npm run test:perf` 2 · `tsc --noEmit`, Biome, Prettier all clean.

## Follow-ups filed, not fixed

- **#60** — the cumulative backoff-cap bail-out (`roka.ts:325`) is a **fourth** bound of this same class that the new guard does not model: `liveMaxRetries >= 5` can never exceed 5 attempts at *any* `turnDeadlineMs`, yet passes the guard silently. Not blocking; shipped value is 2.
- **#59** — surfacing the HTTP status in the log `marker` (deferred: wrong file for this PR's scope fence, and it brushes `decisions.md:64`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for transient gateway timeout responses by enabling retries for HTTP 504 errors.
  * Updated retry-window validation to account for all configured attempts and backoff delays.
  * Increased the default turn deadline to 63 seconds to support the full retry budget.

* **Tests**
  * Expanded coverage for timeout classification, retry timing, and deadline warnings.


AlaskanTuna:
Thanks — the finding is correct, but I'm declining the proposed patch and folding a complete fix into #60 instead. Reasoning, so it's on the record:

**The finding is real.** `roka.ts:324` clamps each delay to the *remaining* cumulative budget (`Math.max(0, retryBackoffCapMs - retryLatencyMs)`), so total backoff per turn is bounded by `retryBackoffCapMs`. `requiredMs` sums **per-delay** caps, which can exceed it. Per-delay sums by retry count: 1000 → 3000 → 7000 → **15000** → 27000, against a 12000 cap. First divergence is `liveMaxRetries: 4`, where `requiredMs` is 115000 but the true need is 112000 — a genuine false warning at deadlines 113000–114999.

**The patch is incomplete in the other direction.** Capping the aggregate at `retryBackoffCapMs` gives `requiredMs = 132000` at `liveMaxRetries: 5`. But that config can never reach 6 attempts at *any* deadline: delays run 1000, 2000, 4000, then 5000 (clamped by the remaining budget) — `retryLatencyMs` reaches exactly 12000 after four retries, so the next `delayMs` is `min(computeBackoff(4), 0)` = 0 and `roka.ts:325`'s `delayMs <= 0 && retryLatencyMs >= retryBackoffCapMs` returns the fallback. Five attempts, not six.

So the patch would go **silent** on that config, whereas the current formula at least warns. It trades a false positive at R=4 for a false negative at R>=5 — and that false negative is precisely what QA independently filed as #60 while reviewing this PR.

**Both are the same missing model:** the guard doesn't derive how many attempts the cumulative cap actually permits. The complete fix models both bounds — require the deadline to fit the *achievable* attempt count, and warn separately when achievable attempts < `liveMaxRetries + 1`. That is #60's scope, and I've updated it to carry both directions rather than patching half of it here.

**Not blocking this merge:** at the shipped `liveMaxRetries: 2` the per-delay sum is 3000, well under the 12000 cap, so `requiredMs = 63000` is exact and this PR's fix is correct and complete for every value currently in `config.yml`. The divergence is unreachable below `liveMaxRetries: 4`.

PR #62, opened 2026-07-31
fix(config): remove a dead key and validate three config footguns

### LINKED ISSUE #54: chore(config): remove dead gemini.baseRetryDelay key
`gemini.baseRetryDelay` is assigned, documented, and configurable, but **nothing in `src/` reads it**.

Found by PL while classifying bounds for #48 (there was no consumer to justify a bound from).

**Surface:**
- `src/config.ts:24` — declared in the `YamlConfig` interface
- `src/config.ts:140` — `baseRetryDelay: yaml.gemini?.baseRetryDelay ?? 2000`
- `config.yml:12` — set to 2000
- `docs/README.md:342` — documented as a tunable
- `src/agent/__tests__/reliability.eval.test.ts:45` — mocked at `0`

Grep for reads of `config.gemini.baseRetryDelay` in `src/` returns nothing. The live retry-backoff path uses `gemini.retryBackoffBaseMs` and `gemini.retryBackoffCapMs` instead (bounded by #44).

**Why it wasn't deleted in #48:** that phase's scope fence was `src/config.ts` and its test file. Deleting this key touches `config.yml` and `docs/README.md` too, and #48 was explicitly "add validation, not retune config". It was bounded at `min: 0` there instead — one line, harmless, and correct if a consumer ever returns.

**Work:** delete the five references above, confirm no consumer appears, and drop the `NUMERIC_BOUNDS` row #48 added. Check whether the test mock at `reliability.eval.test.ts:45` was doing anything real or was cargo-culted alongside the live retry keys.

Low priority — dead config costs a line of confusion, not correctness.

### LINKED ISSUE #55: fix(config): cross-key config invariants are unvalidated (reminders sweep, contextSize vs bufferSize)
Two cross-key relationships are silently breakable. Neither is expressible as a single-key `min`/`max` bound, so #48 deliberately left both out of `NUMERIC_BOUNDS` — they are new behaviour, not validation of the 18.

Found by PL while classifying bounds for #48.

## (a) `reminders.staleThresholdMs` must be >= `reminders.checkIntervalMs`

If the stale threshold is shorter than the check interval, a reminder can be swept as stale **before the next tick can ever deliver it**. Both keys are individually valid at any positive value, so `min: 1` on each does not express the relationship.

Defaults today (`checkIntervalMs: 5000`, `staleThresholdMs: 300000`) satisfy it by a wide margin — this is a footgun for an operator who tunes one without the other, not a live bug.

## (b) `memory.contextSize` > `memory.bufferSize` is silently ineffective

Asking for more context than the buffer holds is harmless but does nothing, and nothing says so. This is the direct analogue of the `extractionInterval`/`bufferSize` clamp already at `src/config.ts:123`.

Note the related trap #48 documented: `contextSize: 0` does **not** mean "no context" — `roka.ts:750` is `.slice(-config.memory.contextSize)`, and `.slice(-0)` is `.slice(0)`, so it injects the *entire* buffer. #48 bounds it `min: 1` for exactly that reason.

## Suggested approach

`src/config.ts:265-279` already carries three cross-key **warnings** (not throws) for relationships like this, so a fourth and fifth would be idiomatic and consistent. Warn rather than throw — both cases are misconfigurations that degrade behaviour rather than break startup, and the existing precedent treats that class as warn-only.

Low priority.

### LINKED ISSUE #56: fix(config): discord.maxMessageLength has no upper bound against Discord's 2000-char hard limit
`discord.maxMessageLength` is bounded `min: 1` (added by #44) but has **no `max`**. Discord's hard limit is 2000 characters, so any configured value above that produces messages the API rejects at send time.

Found by PL while classifying bounds for #48. Not fixed there because it is a **pre-existing** `NUMERIC_BOUNDS` row, not one of that issue's 18 YAML-only keys — changing it would have been scope creep.

**Current state:** `config.yml:33` sets 1500, comfortably under the limit, so this is a footgun rather than a live bug.

**Suggested fix:** add `max: 2000` to the existing row in `src/config.ts`'s `NUMERIC_BOUNDS` table, and extend the table-driven test accordingly. The table's doc comment already establishes that `max` is for values where anything above is "definitionally meaningless" — a message length Discord will always reject qualifies.

Worth confirming the exact limit against current Discord API docs before pinning the number; 2000 is the long-standing non-nitro limit for regular messages, but embeds and nitro have different ceilings and `splitResponse` may already handle chunking in a way that changes what the right cap is.

Low priority.

AlaskanTuna:
Closes #54. Closes #55. Closes #56.

Three config follow-ups Phase 24 deliberately deferred at Gate 1. All warn-only or bounds-only — no behaviour change at any shipped value.

## #54 — delete the dead `gemini.baseRetryDelay`

Eight references, **all inert**: interface member, `?? 2000` assignment, and its own `NUMERIC_BOUNDS` row (`src/config.ts`); `config.yml`; `docs/README.md`; a `toBe(2000)` assertion and an `EXPECTED_BOUNDS` row (`config.test.ts`); a mock (`reliability.eval.test.ts`).

QA confirmed the mock was **provably** inert rather than merely unused: `vi.mock('../../config.js', () => ({config: {…}}))` is a *factory*, so the real module never runs and the factory doesn't even export `NUMERIC_BOUNDS`.

`retryBackoffBaseMs` — the live key with the same job — is untouched across all 28 of its references including its bounds row. The two sitting side by side was the entire confusion this removes.

## #55 — two cross-key invariants, as warnings

Added beside the three existing `logger.warn` blocks. **Neither throws and neither clamps.**

- `reminders.staleThresholdMs <= reminders.checkIntervalMs` — a reminder can be swept stale before any tick can deliver it.
- `memory.contextSize > memory.bufferSize` — silently ineffective, because `.slice(-N)` returns the whole array once `N` exceeds the buffer.

No throw, per #58's Gate 1 reasoning: a throw at startup on a Pi under `restart: unless-stopped` is a crash-loop, and both cases degrade behaviour rather than break it. No clamp on `contextSize` either — unlike the `extractionInterval` clamp at `config.ts:123`, a clamp here would change nothing observable, and `config.memory.contextSize` must keep returning the configured value.

### The operator is `<=`, and the reason is worth reading

This shipped as strict `<`, justified in a comment as *"equality is safe because the previous tick already delivered everything due at or before `now - interval`."* QA was asked to check whether that reasoning was **sound**, not just whether the code matched the spec. It isn't:

- **Cold start.** `startReminderScheduler` (`reminderScheduler.ts:11`) is a bare `setInterval` with **no leading call**, so the first tick at `T0 + I` has no predecessor. Reminders persist in SQLite across restarts, and `reminderStore.ts:84` sweeps `due_at < (T0 + I) - stale`; at `stale == I` that threshold is exactly `T0` — so everything that fell due during downtime is swept **without ever being delivered**, which is precisely the harm the warning names.
- **Timer drift.** `setInterval` fires late, never early, so the real gap `G >= I`; whenever `G > I` the window `(T - G, T - I)` is swept undelivered.

Equality is safe only in steady state. Since `main` auto-deploys on every merge, cold start is routine in this deployment rather than exotic — so the check now fires at equality, and the comment states the assumption and cites both source lines instead of asserting safety.

## #56 — bound `discord.maxMessageLength` at 4000, not 2000

**The issue asked for 2000 and the issue is wrong.** It assumed the bot sends through Discord's `content` field. It does not: QA traced all four send sites (`messageCreate.ts:336,340`, `interactionCreate.ts:122,125`) through `buildRokaMessage`, whose payload object literally **has no `content` key** — it sets `flags: MessageFlags.IsComponentsV2`. The binding ceiling is the **4000-character budget shared across all `TextDisplay` components**.

`max: 2000` would have encoded a fact that is false for this bot and rejected configs that work. The row carries a comment citing `messageBuilder.ts:57-60` so a future reader doesn't "correct" it back.

`config.yml`'s value stays at **1500** — this adds validation, not tuning. Only its (previously wrong) comment changed.

## Mutation pinning (2026-07-29 convention)

Every probe produced its predicted failure, then was restored:

- `discord.maxMessageLength`: production bound alone → 1 failure (`expected 4000 to be undefined`); `EXPECTED_BOUNDS` row alone → 2 failures. The paired-table pin is real in both directions.
- Cross-key boundaries: flipping `<`→`<=` and `>`→`>=` failed exactly the two equality tests; after the M-1 change, flipping `<=` back to `<` failed exactly the new equality test.
- QA ran its probes against SHA-256-verified snapshots and restored both files byte-identically.

Every invariant has a **firing case and a silent case immediately adjacent to the threshold**, so an off-by-one is caught in both directions.

## Verification

`npm test` **760 passed / 57 files** (753 baseline: +4 #55, +2 #56, +1 the new adjacent silent case, +0 net #54) · `npm run test:perf` 2 · `tsc --noEmit`, Biome, Prettier all clean.

## Deferred, not fixed

The tool-disclosure `TextDisplay` (`messageBuilder.ts:46-55`) shares the same 4000 budget, so a `maxMessageLength` near 4000 **plus** a fired tool can exceed it. Unreachable at the shipped 1500. Filed separately.

## Summary by CodeRabbit

- **New Features**
  - Added startup warnings when reminder settings could delay delivery.
  - Added warnings when configured memory sizes exceed the available passive context buffer.

- **Bug Fixes**
  - Improved configuration validation by enforcing a maximum Discord message length of 4,000 characters.
  - Updated message-length guidance to reflect the shared Components V2 text budget.

- **Documentation**
  - Removed the obsolete Gemini retry-delay setting from configuration guidance.


AlaskanTuna:
Adopted the wrap; declined the suggested wording.

The second line of the proposed fix — `The 2000-character content limit in src/discord/messageBuilder.ts is unused` — misplaces the fact. The 2000-character limit is **Discord's limit on the `content` field**; it is not defined in `messageBuilder.ts`. That file is the *evidence* that this bot never uses `content` — it sets `flags: MessageFlags.IsComponentsV2` and its payload object has no `content` key at all.

Phrasing it that way would defeat the comment's entire purpose, which is to stop a future reader from "correcting" the bound back to 2000. Wrapped to two lines (100 and 79 chars) keeping the fact intact, and dropped the em-dash for ASCII while there.

Also kept the word **shared** deliberately: the 4000 budget spans all `TextDisplay` components, which is why the tool-footer residual (#63) exists.

PR #84, opened 2026-08-18
chore: pin the passive-extraction guild gate and unstale the PRD's split threshold

### LINKED ISSUE #76: docs: prd.md's 2000-char split threshold is stale — the ceiling is derived, not a literal
Found while doing Phase 27's documentation task (#29), flagged rather than fixed in passing per this repo's surgical-change discipline.

## The Drift

`docs/prd.md:61` states that a reply is **"split if >2000 chars"**.

That is wrong. This bot sends every reply through Components V2 — both reply paths set `MessageFlags.IsComponentsV2` and never use `content` — so the governing limit is the shared `TextDisplay` budget of 4000, not `content`'s 2000 (`docs/decisions.md` 2026-07-30, from #56).

It is wrong in a second, subtler way. The real ceiling is not 4000 either: `discord.maxMessageLength` is bounded in code at `4000 - MAX_TOOL_FOOTER_CHARS`, derived from `TOOL_USAGE_LABELS` (`docs/decisions.md` 2026-08-01, from #63). A message at exactly 4000 becomes unsendable the moment any tool fires and the footer is appended.

## Why This Line Specifically Is Worth Care

This is the **third** documented instance of the same value being wrong in this repo — #56 asked for `max: 2000` on the same misunderstanding, and PR #62 then shipped `max: 4000`, which #63 had to lower to 3878 one day later. Each correction was made against the previous wrong number rather than against the code.

So the fix is **not** to write 3878 into the PRD. `docs/decisions.md` 2026-07-31: **documentation states that a constraint exists and where it is enforced, never its value** — the value lives in the constant a test checks, and this particular value is *derived*, meaning any number written in prose is stale the moment a tool label changes.

## Suggested Fix

Reword `docs/prd.md:61` to state that long replies are split, that the ceiling is the Components V2 `TextDisplay` budget minus the tool footer, and that it is enforced by the bound on `discord.maxMessageLength` in `src/config.ts` — naming the enforcement point, not the number. One line.

While in there, check whether any sibling line in `docs/prd.md` carries the same assumption; a single grep for `2000`, `4000`, and `chars` across the live docs would settle it. Do not widen beyond that.

Per this repo's standing rule, confirm against the code rather than this issue text: read the two reply paths for the `IsComponentsV2` flag and absence of `content`, and read the `maxMessageLength` bound in `src/config.ts` for how the ceiling is derived.

### LINKED ISSUE #77: test(discord): the passive-extraction guild gate is unpinned — probe P5 does not fire
Found by mutation probe **P5** during Phase 27 (#29). The probe was predicted in advance not to fire, run honestly anyway, and did not fire — which is the finding.

## The Gap

`src/discord/events/messageCreate.ts` gates passive claim extraction (and the legacy extractor) on `message.guild`, so neither can run outside a guild. That is correct and deliberate: user-install contexts deliver interactions only, with no gateway message stream, and passive extraction outside a guild would write memory nobody asked for.

**Nothing asserts it.** Removing the `message.guild &&` conjunct at the passive-extraction gate leaves all 31 tests in `src/discord/events/__tests__/messageCreate.test.ts` green.

The fact holds by accident, and `docs/decisions.md` records the standing rule that a claim nobody can fail a build on will eventually be false — the root cause of most of Phase 26, where one rule recurred three times (#41 → #43 → #58) because it lived in a comment.

## Why It Was Not Written

Phase 27's plan assigned this assertion to task 181 (task 183 explicitly deferred to it: *"Items 2 and 3 are assertions on task 181's file, so they belong to task 181"*). Task 181 was then carved out and shipped standalone as the DM-tenant privacy hotfix (PR #73), scoped to the tenant derivation only. The assertion fell through the seam between the two.

## The Missing Assertion, Precisely

In `src/discord/events/__tests__/messageCreate.test.ts`: a case built with the existing `createMessage({ guild: null })` helper, with `isMonitored` mocked `true`, asserting `mocks.addToPassiveBuffer` was **not** called.

The mock already exists in that file. Only the assertion is missing. Consider covering the sibling gates in the same pass — the legacy extractor and the emoji-reaction path are guarded identically and are equally unpinned.

## Acceptance

Pin it by mutation, per `docs/decisions.md` 2026-07-29: remove the `message.guild &&` conjunct, confirm the new assertion fails, restore. A test that passes before the guard is removed is not a regression test.

**One structural warning, learned four times in this repo now** (Phase 26's P1d/P2c/P3b, and twice during Phase 27): Vitest stops at the first failing assertion in a block, so a coarse assertion ahead of a specific one masks it and the probe fires on the wrong thing. Give this its own `it` rather than appending it to an existing one.

## Scope

Test-only. No production change — the guard is already correct. Do not "fix" `messageCreate.ts`.

Per this repo's standing rule, confirm against the source rather than this issue text: read the passive-extraction and legacy-extractor call sites in `src/discord/events/messageCreate.ts` for the actual guard expression, and `src/discord/events/__tests__/messageCreate.test.ts` for the `createMessage` helper's current signature.

AlaskanTuna:
Closes #77
Closes #76

Two Tier-1 items, both premise-verified against source before implementing. **No production source is touched** — `src/discord/events/messageCreate.ts` is byte-identical to `main`.

## #77 — Pin the passive-extraction guild gate

`messageCreate.ts:131` gates passive claim extraction on `message.guild`. That was correct and deliberate but nothing asserted it: `mocks.addToPassiveBuffer` was declared and wired in the test file with **zero assertions anywhere**.

Adds three `it` blocks — one per sink inside the guard (`addToPassiveBuffer`, `maybeExtractFromBuffer`, `enqueueAndSchedule`), each in its own block per the convention that a coarse assertion ahead of a specific one masks it.

**Mutation probe.** Removing the `message.guild &&` conjunct at `:131` fails all three and leaves the 31 pre-existing tests green; each failure text matched its prediction. Restored from a filesystem snapshot, md5 verified identical.

Two corrections to the issue text, both recorded:
- The test file is `src/discord/__tests__/messageCreate.test.ts`, not `src/discord/events/__tests__/…` — that path does not exist.
- `createMessage`'s `guild` parameter already defaults to `null`, so `{ guild: null }` alone would be vacuous. The load-bearing ingredient is the enclosing describe's `isMonitored → true`.

**Known limitation, deliberately not closed.** The sibling gate at `:358` is not pinned and cannot be by a not-called assertion — its first statement dereferences `message.guild.members.me`, so removing the conjunct throws a `TypeError` that the surrounding `try/catch` swallows before any sink is reached. Closing it needs a production change, which #77 forbids.

## #76 — The PRD's split threshold

`docs/prd.md` claimed replies split at 2000 chars. Both reply paths set `MessageFlags.IsComponentsV2` and never use `content`, so the governing budget is `TextDisplay`'s, and the real ceiling is that minus the tool footer.

Per the standing rule that docs state where a constraint is enforced and never its value, both lines now name the bound on `discord.maxMessageLength` in `src/config.ts` and its pin in `messageBuilder.test.ts` — no number restated. This is the third time this value has been documented wrong, each correction made against the previous wrong number.

**Wider than the issue stated:** the stale claim was on **two** lines, `:61` and `:159`. The issue named only `:61`; its own instruction to grep surfaced the second. Post-edit sweep of all committed docs returns one hit, a SQL column name.

## Gates

| Gate | Result |
| ---- | ------ |
| `npx vitest run` | 814 / 61 (was 811 / 61) |
| `npm run test:perf` | 2 / 1 |
| `npx tsc --noEmit` | clean |
| `biome check .` | clean, 179 files |
| `prettier --check .` | clean |

Zero paid Gemini calls; `npm run test:live` not run and not owed — no prompt layer, tool description, or assembly path is touched.

## Summary by CodeRabbit

- **New Features**
  - Long responses are now split into multiple messages when they exceed the platform’s message length limit.
  - Message sizing accounts for enhanced display components and accompanying tool information.

- **Bug Fixes**
  - Direct messages no longer trigger passive buffering or unintended content-extraction workflows.

- **Documentation**
  - Updated workflow guidance to describe response splitting and message length limits.


AlaskanTuna:
@coderabbitai review

PR #85, opened 2026-08-18
docs(discord): record the guild install scope and permission derivation

### LINKED ISSUE #75: docs(discord): record the guild install scope and permission requirements, or the derivation behind them
Surfaced while enabling User Install for #29 (session 2026-08-01). The Portal's Guild Install default settings previously requested only `applications.commands` — **no `bot` scope** — so anyone installing Rokabot through the Discord-provided link would have received an app with no bot member: no gateway events, therefore no mention/reply/name-keyword triggers and no reactions. The existing server install predates that link and was made with a hand-built OAuth URL, which is why nothing was ever observed to be wrong.

That is now fixed in the Portal. This issue is about the fact that **nothing in the repo records what Rokabot needs, or why** — so the next person to touch it re-derives it, or doesn't.

## What Was Derived, and From Where

The guild permission set is not arbitrary; every bit traces to a Discord API call in `src/`:

| Permission | Required By |
| ---------- | ----------- |
| View Channels | the `messageCreate` path existing at all |
| Send Messages | `src/discord/events/messageCreate.ts:306,340`, `src/discord/events/gameCommands.ts:50,72,84` |
| Read Message History | `src/discord/events/messageCreate.ts:114` — fetching the referenced message on a reply trigger |
| Add Reactions | `src/discord/events/messageCreate.ts:122` |
| Attach Files | `src/discord/events/stats/views.ts:442-474` — the `/stats` chart PNGs |
| Embed Links | nothing today; Components V2 messages carry no embeds and bare URLs stay clickable without it. Pre-granted for #19's hyperlink-footnote work, which would introduce masked links. |

Reminder delivery to users is a DM (`src/discord/reminderScheduler.ts:25,42`) and needs no guild permission.

## The Design Tension to Resolve

`docs/decisions.md` 2026-07-31: **documentation states that a constraint exists and where it is enforced, never its value** — because a value nothing can fail a build on will eventually be false. That rule was learned the expensive way (four docs carried a prompt budget roughly 2× wrong).

A permission set is a value. But unlike most values in this repo it lives in the **Discord Developer Portal**, outside the tree, where no test can reach it. So the usual counter-move — tie an assertion to the thing it models — is not straightforwardly available.

Three options, in rough order of preference:

1. **Document the derivation, not the number.** `docs/runbook.md` states that a guild install requires the permissions implied by Rokabot's Discord API surface, and points at the call sites above. Anyone can re-derive the current set by reading them; nothing goes stale because nothing is pinned. Closest in spirit to how #65 resolved `AGENTS.md`.
2. **Document the set AND state plainly that it has no build-time pin**, the way #65's outcome was recorded — honest, useful, and explicitly accepts the rot risk rather than pretending it away.
3. **Attempt an actual pin.** A test that walks `src/` for the Discord operations used and asserts they match a declared list. Consider this carefully before adopting: it is a brittle static analysis over a moving surface, and it would pin *our list against our grep*, not against the Portal — which is where the value actually lives and where the drift would actually happen. That is the "passes for the wrong reason" failure mode (`docs/decisions.md` 2026-07-29). **I would expect this option to be rejected**, but it deserves an explicit rejection rather than silent omission.

Whichever is chosen, the **`bot` scope requirement is the load-bearing part** — it is the thing whose absence produced a silently half-broken install, and it is worth stating on its own rather than folding into a permission table.

## Scope

Documentation only. No code change. `docs/runbook.md` is committed (unlike the local-only pm-workflow artifacts), so this ships as a normal PR.

Per this repo's standing rule, confirm every call-site citation above against the source rather than trusting this table — it was assembled by grep during a session, not by an audit.

AlaskanTuna:
Closes #75

Documentation only, no code change. Adds a **Discord Install Requirements** section to `docs/runbook.md`.

## Approach — the issue's Option 1, with Option 2's honesty

The issue laid out three options and pre-argued that Option 3 (a grep-based test) should be rejected. Agreed, and the rejection is now recorded *in the doc* rather than silently omitted: such a test would pin our list against our own grep, not against the Portal — which is where the value lives and where drift actually happens. That is the "passes for the wrong reason" failure mode.

So the section documents the **derivation** (operation → permission), names the files, and states plainly that no build-time pin exists because the Portal is outside the tree. The `bot` scope requirement is stated on its own, as the load-bearing part — its absence is what produced a silently half-broken install.

**Line numbers are deliberately omitted.** They rot, and this proved it during the audit — see below.

## Premise audit — the issue's table was assembled by grep, and it shows

The issue asked for its citations to be re-verified rather than trusted. Doing that turned up two defects:

- **`messageCreate.ts:306` is not a send site.** The actual reply/send call sites in that file are `:289` (busy), `:300` (decline), `:336` (main reply), `:340` (chunk send) and `:376` (error). The issue cited `:306,340`, so it was both wrong and incomplete.
- **The reminder claim is backwards.** The issue states *"Reminder delivery to users is a DM … and needs no guild permission"*, citing `reminderScheduler.ts:25,42`. Those two lines are the **fallback**. The primary path is `channel.send(...)` at `:17` — the module's own docstring reads "delivers due reminders to Discord **channels**". Reminders therefore **do** need guild Send Messages, and the doc says so.

Everything else verified exact: `gameCommands.ts:50,72,84`, `messageCreate.ts:114` (history fetch), `messageCreate.ts:122` (react), `stats/views.ts:442-474` (five `AttachmentBuilder` sites), and "Embed Links: nothing today" — confirmed, zero `EmbedBuilder`/`embeds:` in non-test source.

That is the fifth issue in this repo found to carry a backwards or drifted premise, and the direct reason this section cites files and operations instead of line numbers.

## Gates

`npx tsc --noEmit` clean · `biome check .` clean (179 files) · `npx vitest run` 811 / 61 · `prettier --check .` clean (sole warning is the git-excluded `docs/.kickoff-next-session.md`, pre-existing).

Zero paid Gemini calls.


AlaskanTuna:
@coderabbitai review

Issue #25, opened 2026-07-23
feat(agent): semantic memory retrieval via embeddings + guild-wide memory search


AlaskanTuna:
## Motivation

Live testing surfaced two structural gaps in the claims memory system:

1. **Recall is recency-keyed, not relevance-keyed.** `recall_user` returns the 15 freshest facts (`MAX_RECALLED_FACTS`, PR #24). For members with many claims, a relevant-but-older fact (e.g. `works_as`) falls outside the window while recent trivia fills it.
2. **Topic-keyed questions are unanswerable.** "Who played this game before in this server?" fails because every memory tool is keyed on a *person* — there is no tool shaped like "search the guild's memory by subject." FTS5 cannot bridge the meaning gap ("played this game" vs a claim stored as `plays: <title>`).

Embedding-based retrieval closes both: lexical search matches words, vector search matches meaning.

## Proposed Scope

1. **Embedding pipeline** — embed claim values at write time using the Gemini embedding model (separate free-tier quota, ~1K RPD; well within extraction volume). 768-dim (Matryoshka-truncated), stored via `sqlite-vec` in the existing SQLite database. One-time backfill job for existing claims. Strictly guild-scoped: vectors filtered by `guild_id` before any similarity ranking, never ranked across guilds (per the no-crossing memory contract).
2. **Relevance-ranked `recall_user`** — rank the subject's claims by cosine similarity to the current message, blended with freshness, instead of freshness alone. `MAX_RECALLED_FACTS = 15` stays; what fills the slots gets smarter.
3. **New tool: `search_memories`** — topic-keyed, guild-wide semantic search over claim values ("who here is into cooking?"). This is the new agentic capability. Privacy rules need design at Gate 1: active + `needs_review = 0` claims only, strictly per-guild; consider a salience floor, since it surfaces facts about members who were not asked about by name.
4. **Write-time near-duplicate detection** — if a new claim lands within a similarity threshold of an existing claim for the same subject, route to the supersede path instead of accumulating paraphrase duplicates (`hobby: brewing coffee` vs `likes: making coffee`). Keeps claim counts honest for /stats "Who She Knows Best".

## Constraints & Failure Modes

- `sqlite-vec` must build/run on ARM64 Alpine (Docker, RPi 5) — verify in-container, not only on the dev machine.
- Embed call sits in the hot path for memory-touching turns (~100–200 ms added): on embed failure, fall back to current freshest-first behavior — never block or deflect a reply over retrieval.
[...]

PR #86, opened 2026-08-18
fix(agent): rank recalled claims by salience instead of recency alone

AlaskanTuna:
Partially addresses #25 (item 1). **Does not close it** — items 2–4 remain.

## What the premise check found

The session handoff predicted that `recall_user` already blended relevance and that #25's first two items would evaporate. **The truth is the inverse**, and it changes #25's shape considerably:

- **Passive prompt injection already blends relevance.** `src/agent/memory/retriever.ts` → `scoreClaim` combines an FTS `bm25()` hit (+1.5), `salience × sourceWeight × 2`, confidence, `recency × 0.5`, pinned (+1) and routed predicates (+1).
- **The `recall_user` tool did not.** `recallUser.ts` re-sorted `[...claims].sort((l, r) => r.lastSeenAt - l.lastSeenAt)`, **discarding** the `pinned DESC, salience DESC, last_seen_at DESC, id DESC` ordering that `getActiveClaims` had already applied, then took the first 15.

So #25 item 1 is real but confined to the tool path, and the fix is not embeddings — it is deleting the re-sort. That is what this PR does.

## Why it matters concretely

`remember_user` writes claims with `sourceKind: 'explicit'` (salience `0.5`); the passive extractor writes `sourceKind: 'passive'` (salience `0.375`). Because the re-sort ignored salience, **a fact the user explicitly asked Roka to remember could be pushed out of the 15-slot recall window by 15 fresher passively-extracted trivia.** That is now the regression test.

## Mutation probe

Restoring the recency re-sort fails the new test with exactly the predicted message:

```
FAIL  keeps an explicitly remembered claim in the window against fresher passive trivia
AssertionError: expected 'likes: thing-14, likes: thing-13, lik…' to contain 'general_occupation: shrine caretaker'
```

Restored by `cp` from a filesystem snapshot; md5 `d04a65dd388080c98b76cc409689a937` before and after. `git checkout --` was not used.

## Is any existing assertion weakened?

**No — nothing was edited to fit.** The existing `'recalls freshest claims first and caps the merged list at 15…'` test is untouched and passes under both orderings: every claim in it shares one predicate and one source kind, so their salience is identical and the ordering falls through to `last_seen_at DESC` either way. It could never have distinguished the two behaviours. The suite is **strictly stronger** — one added assertion, none removed or loosened.

## Premise correction worth recording separately

`baseSalience` is **hardcoded to `0.5` for every predicate** — the `predicate()` factory in `src/agent/memory/predicates.ts` sets it unconditionally, so `baseSalienceOf()` is a constant. Predicate importance is therefore **not modelled anywhere today**, and salience varies only by source kind and by the `+0.02` accumulated per re-observation.

This matters for #25's remaining scope: the issue's motivating example ("a relevant-but-older `works_as` fact falls outside the window") cannot be fixed by salience as currently defined, because `general_occupation` and `likes` carry identical base salience. Either predicate weighting becomes real, or relevance has to come from the message — which `recall_user` does not currently receive.

## Scope

`recall_user`'s output ordering only. No prompt layer, no tool description, no ADK path — so no live re-measurement is owed and zero paid Gemini calls were made. Behaviour change is non-breaking: the return shape, the 15-cap, the claims-before-legacy-facts merge order and the `touchRecalled` set are all unchanged.

## Gates

`npx vitest run` **815 / 61** (was 814) · `npm run test:perf` 2 / 1 · `npx tsc --noEmit` clean · `biome check .` clean (179 files) · `prettier --check .` clean apart from the git-excluded `docs/.kickoff-next-session.md`.


AlaskanTuna:
@coderabbitai review

PR #87, opened 2026-08-18
test(agent): pin ADK's stateDelta propagation into toolContext.state

### LINKED ISSUE #72: test(agent): pin ADK's stateDelta to toolContext.state propagation with an unpaid contract test
Raised by QA (Codex `gpt-5.6-sol`, xhigh) on the third review pass of the Phase 27 DM-tenant hotfix. Approved as a **documented limitation, not a ship blocker** — filed so the gap is tracked rather than forgotten.

## The Gap

The hotfix added `src/discord/__tests__/interactionCreate.memory.test.ts`, which pins the memory tenant through the entire repo-owned path: interaction handler → `generateResponse` → the request object actually handed to `runner.runAsync`. That boundary is mutation-pinned; QA confirmed it cannot construct a mutation of our tree that reopens cross-DM access while the suite stays green.

What is **not** pinned is the next hop: ADK's propagation of `stateDelta` into `toolContext.state`. The test reconstructs that Map from the captured `stateDelta` rather than observing ADK perform the propagation.

## Why It Is Acceptable Today

QA verified against the installed ADK source that the hop is really implemented: `Runner` applies `stateDelta` via `sessionService.appendEvent`, `CallbackContext` derives state from `invocationContext.session.state`, and `ToolContext` inherits that context. The dependency is lockfile-pinned. Combined with the mutation-pinned request boundary and the fail-closed guards on missing/`'global'` tenant state, the risk of an undetected regression is low.

But it is a real hole: an ADK upgrade that changed this propagation would not fail any test in this repository.

## The Work — a Cheap Unpaid Contract Test

QA's finding, verbatim:

> The programmer did miss a cheap unpaid test: a scripted `BaseLlm`, or a mocked `GoogleGenAI` returning a deterministic function call, could drive the real ADK `Runner` into the actual memory tool without reimplementing state plumbing. No existing repository test pins that result; the live gate exercises real tool calls but scores tool firing rather than the tenant received.

So: drive the **real** ADK `Runner` with a scripted model that deterministically emits a function call, and assert the tenant the memory tool actually receives in `toolContext.state`. No network, no paid calls, no reimplementation of ADK's state plumbing — which is the trap the current test deliberately avoided, since pinning our own re-implementation would be worse than not pinning at all.

## Constraints

- **Zero paid calls.** The model must be scripted or mocked. `npm run test:live` is a separate opt-in gate and is not the answer here.
- **Do not reimplement ADK's state plumbing in the test.** If the only reachable design ends up asserting a re-implementation, that is a finding — report it rather than shipping a test that pins the wrong thing.
- Pin it by mutation like everything else in this area (`docs/decisions.md` 2026-07-29): the new test must be observed failing when the propagation it models is broken.

## Why This Matters Beyond One Test

The value is upgrade safety. This assertion is what would catch an ADK version bump silently changing how session state reaches tools — a class of failure that is invisible to every other gate in the repo, including the live one, which scores tool *firing* rather than the tenant received.

Per this repo's standing rule, confirm against the code rather than this issue text: read `src/agent/roka.ts` around the `testRunTurn` early return and the `runner.runAsync` request construction below it, and `src/discord/__tests__/interactionCreate.memory.test.ts` for what the current capturing-`Runner` test does and does not observe.

AlaskanTuna:
Closes #72.

## What

Adds `src/agent/__tests__/adkToolContextState.contract.test.ts` — a contract test against **ADK itself**, driving the real `Runner` with a scripted `BaseLlm` that deterministically emits a `remember_user` function call, and asserting the tenant the real memory tool receives in `toolContext.state`.

No production code changed. No network, no paid calls.

## Why

`interactionCreate.memory.test.ts` pins the request this repo builds and stops at `runner.runAsync`. It reconstructs `toolContext.state` from the captured `stateDelta` (`new Map(Object.entries(stateDelta))`) rather than observing ADK propagate it. An ADK upgrade that changed that propagation failed no test in this repo.

## Mutation probe

ADK's propagation point in `node_modules/@google/adk/dist/esm/index.js` — `Runner.runAsync`'s `actions: r ? F({ stateDelta: r }) : void 0` — was neutered to `actions: void 0`, simulating an ADK version that stopped applying the caller's `stateDelta`.

Both new assertions failed:

```
1. stores the fact under the tenant the runAsync stateDelta named
   AssertionError: expected [] to deeply equal [ { key: 'favorite_anime', ...(1) } ]

2. reaches the tool with usable tenant state rather than tripping its fail-closed guard
   AssertionError: expected [ [ 'remember_user', false ] ] to deeply equal [ [ 'remember_user', true ] ]
```

`interactionCreate.memory.test.ts` **stayed green under the same break** — which is exactly the hole #72 describes, now closed. `node_modules` was restored from a byte-verified backup (md5 `d575b84ffd012dfc28ffd3d9cb7856e9`).

## Gates

- `npx vitest run` — 813 passed / 62 files, exit 0
- `biome check .` — 180 files, no fixes


AlaskanTuna:
@coderabbitai review

PR #88, opened 2026-08-18
fix(agent): filter name resolution by tenant instead of preferring it

### LINKED ISSUE #70: fix(agent): name resolution prefers rather than filters by tenant, leaving an existence oracle
Raised by QA (Codex `gpt-5.6-sol`, xhigh) during review of the Phase 27 DM-tenant hotfix. Marked **LOW** and explicitly **pre-existing — not caused by that diff**. Filed rather than fixed so the analysis survives.

## The Finding, Verbatim

> `src/storage/userNames.ts:59` — LOW — PRE-EXISTING CONDITION, not caused by this diff: the non-global query only prefers current-tenant users; it does not filter them. A globally known user remains resolvable from another DM, exposing the distinction between "known, but no notes" and "unknown name." No fact value crosses tenants — filter candidates by current-tenant claims, legacy facts, or response activity.

## What This Means

`findUserByName(name, guildId)` has two branches (`src/storage/userNames.ts:47-68`). The non-`'global'` branch does **not** filter by tenant — it `ORDER BY`s an `EXISTS` preference over `memory_claim` and `response_events`, then `LIMIT 1`. So a user known anywhere is still resolvable from any tenant.

Post-hotfix, `recall_user` for that user under a foreign tenant returns `factCount === 0`, because both `getActiveClaims` and `getFacts` are queried with the caller's own tenant. **No fact value crosses.** What remains is an existence oracle: Roka answers "I don't know anything about them" (name resolved, no facts) rather than "I don't know anyone by that name here yet" (`src/agent/tools/index.ts:158`), and the difference is observable.

## Severity Rationale

Lower than it first looks: it leaks *that a name is known to the bot somewhere*, not any stored fact. It predates the DM tenant work — the same distinction existed between guilds. It is worth closing because the DM/group-DM surface multiplies the number of tenants from which the question can be asked.

## Suggested Fix

QA's own suggestion: filter candidates by current-tenant claims, legacy facts, or response activity rather than merely preferring them. Note the trade-off before implementing — a hard filter means Roka cannot resolve a name on the very first interaction in a new tenant, before any claim or `response_events` row exists for that user there. Decide deliberately whether the not-found path should be reachable for a genuinely present member.

Per this repo's standing rule, confirm against the code rather than this issue text: read `src/storage/userNames.ts:47-68` for the two branches, `src/agent/tools/index.ts:157-159` for the caller and the two distinguishable messages.

AlaskanTuna:
Closes #70

`findUserByName(name, guildId)`'s non-global branch used `ORDER BY (EXISTS(claims) OR EXISTS(activity)) DESC LIMIT 1` — it **preferred** in-tenant users but never excluded out-of-tenant ones, so a name known anywhere stayed resolvable everywhere. No fact value crossed (both `getActiveClaims` and `getFacts` are queried with the caller's own tenant), but the *distinction* between "known, no notes" and "unknown name" was observable — an existence oracle, and the DM/group-DM surface multiplied the number of tenants it could be asked from.

Now a `WHERE` filter over three tenant-evidence sources: claims, legacy facts, and response activity.

## Premise verification

- Confirmed at `src/storage/userNames.ts` — the branch really did prefer rather than filter.
- Confirmed the two distinguishable messages at `src/agent/tools/index.ts`: `"I don't know anyone by that name here yet."` on an unresolved name vs `"I don't have any notes about this person yet."` from `recallUser` on an empty result.
- **`findUserByName` has exactly one production caller** — `recall_user` with a `user_name` argument. `remember_user` does **not** use it; it only ever writes about the current speaker via `_userId`. So the blast radius is one path, which is smaller than the issue implies.

## The trade-off the issue asked to be weighed, resolved

The issue warns that a hard filter means a name cannot resolve on the first interaction in a new tenant. That is true, and it costs less than it sounds: if a user has no claims in the tenant, `recallUser` returns `factCount: 0` regardless. **The information delivered is identical either way — only the wording changes**, from "I don't have any notes about this person yet" to "I don't know anyone by that name here yet", and in a tenant where nothing is known the second sentence is the more accurate one. That is why this ships filtered rather than staying preferred.

The filter also improves name-collision handling: two members sharing a display name where neither is present in the tenant now resolve to nothing rather than to an arbitrary `LIMIT 1` pick.

## Legacy facts are load-bearing, not polish

The QA finding's suggested fix named "claims, **legacy facts**, or response activity". Including `user_memory` turned out to be required, not optional — **probe B below shows that omitting it breaks a pre-existing test.** A naive two-source filter would have shipped a regression.

## Mutation probes

Snapshot/`cp` restore, md5 `9678a2a790a4f50d40544c3163a0665d` verified identical after each. `git checkout --` not used.

- **Probe A** — revert the `AND (…)` filter to the old `ORDER BY (…) DESC`: fails exactly `does not resolve a globally known name from a tenant the user has no presence in`, 14 others green.
- **Probe B** — delete the `user_memory` clause: fails **two** tests — the new `resolves a name backed only by a legacy fact in the current tenant` *and* the pre-existing `recalls a resolved user_name through the FunctionTool`.

Four new `it` blocks, one per case (negative + the three evidence sources), each in its own block per the convention that a coarse assertion ahead of a specific one masks it.

## Gates

`npx vitest run` **819 / 61** (was 815) · `npm run test:perf` 2 / 1 · `npx tsc --noEmit` clean · `biome check .` clean (179 files) · `prettier --check .` clean apart from the git-excluded `docs/.kickoff-next-session.md`.

Zero paid Gemini calls; no prompt layer or tool description touched, so no live re-measurement is owed.


AlaskanTuna:
@coderabbitai review

Issue #81, opened 2026-08-05
fix(agent): de-escalate context on safety blocks instead of deflecting after one ineffective retry


AlaskanTuna:
## Summary

When Gemini's non-configurable safety filter rejects a turn, Roka retries **once** with the *same* conversation context plus a steering addendum, then gives up and emits the deflection `"Ehh… let's not get into that one~"`.

Because the trigger lives in the carried-over context rather than the current message, that retry cannot succeed. Every observed safety deflection in `response_events` shows `retries: 1` followed by `outcome: deflection` — a 100% failure rate for the existing mitigation.

## Impact: the failure is self-sustaining

Once prohibited content enters a channel's rolling context (session history + the overheard block from `buildOverheardBlock()`), **every subsequent turn re-sends it**, so replies to entirely innocuous messages are also blocked. Each new message refreshes the session TTL, which keeps the contaminated context alive.

Production evidence — a burst where the channel stayed broken for 11 minutes, then recovered on its own:

```
23:18:15  deflection ┐
23:18:33  deflection │  every new message refreshes session.ttl (500000ms),
23:18:50  deflection │  so the poisoned context never expires
23:19:22  deflection │
23:26:05  deflection │
23:29:19  deflection ┘
          (channel quiet 9m16s)
23:37:39  session idle-timeout → context destroyed
23:38:35  OK — and every turn afterwards
```

Recovery correlates exactly with `session.ttl` expiry, not with anything the bot did. The perverse consequence: the more users talk while she is stuck, the longer she stays stuck.

Base rate across all recorded turns: **16 safety deflections out of 200 responses (8%)**, arriving in bursts (45% and 40% of turns on the two worst days, ~0% otherwise).

## Not fixable via configuration

`config.yml` already sets `safetyThreshold: OFF` for all four adjustable categories (`HARASSMENT`, `HATE_SPEECH`, `SEXUALLY_EXPLICIT`, `DANGEROUS_CONTENT`). These rejections carry `PROHIBITED_CONTENT`, which comes from Google's non-adjustable filter and cannot be relaxed by any safety setting.

The mitigation therefore has to be prompt-side.

## Proposed solution: de-escalation ladder

Replace the single steering retry with a ladder that sheds the riskiest *optional* context one rung at a time, keeping the current user message intact:

```
attempt 0  full prompt
   ↓ blocked
attempt 1  drop overheard block        (other users' raw messages)
   ↓ blocked
[...]

PR #93, opened 2026-08-18
feat(agent): make Roka search the web when she is unsure of a fact

### LINKED ISSUE #19: enhancement: refine web search — footnote links, dialogue delivery, and provider evaluation

AlaskanTuna:
Closes #19 items 4, 5, 6 and 7.

`search_web` fired on **1.7% of production turns** (4 of 236 measured). Asked "what is GPT Astra", Roka answered from pretrained knowledge instead of looking it up. Two things caused that, both self-inflicted: the tool description opened with `FALLBACK: … only when the user explicitly asks about recent news or current events`, and the core prompt's Hard Boundaries told her to deflect anything she "wouldn't know as Roka" — which routed answerable factual questions away from the tool.

## The measurement

Item 6 specifies a should-search / shouldn't-search benchmark as the acceptance gate for item 4. New fixture `tests/harness/tool-trigger/search-web.jsonl`, 6 should-fire / 6 should-not-fire. **No harness code was needed** — the rig scores `result.toolsUsed.includes(header.tool)` and `header.tool` is already generic.

**3 trials × 12 cases: TP=18, FP=0, TN=18, FN=0 — precision 1.000, recall 1.000, zero systematic failures.** Every should-fire case fired 3/3; every should-not-fire case 0/3. No tuning iterations.

Both arbitration cases held: *"what do you know about Mio?"* stayed with `recall_user`, and *"how's the sweets shop doing?"* triggered nothing.

## Why the rubric is positive-only — this is the load-bearing design decision

Item 6 proposes an explicit "never search" list (greetings, banter, opinions, roleplay). **The #19 comment of 2026-07-29 measured that construct backfiring on #39**: a standalone `Never …` rule fixed its worst case outright (12/12 → 0/3), but the one category the rule *enumerated* got worse — general-advice went 3/9 → 3/3 in both attempts. Enumerating a category inside a negative instruction appears to make it salient rather than suppressing it.

So the rubric states only the positive condition and lets the absence of a trigger do the suppressing. That comment is why this measured clean on the first attempt rather than costing an extra 120-call cycle.

Same comment's other lesson applied: structural placement beat wording on #39, so these are standalone bullets rather than clauses appended to existing ones.

## Regression gate — two consecutive green runs, as required

The existing `recall_user` gate, criterion `precision >= 0.80 && recall >= 0.80`:

| Run | TP | FP | TN | FN | Precision | Recall | Accuracy |
| --- | -- | -- | -- | -- | --------- | ------ | -------- |
| 1 | 18 | 2 | 16 | 0 | **0.900** | **1.000** | 0.944 |
| 2 | 18 | 2 | 16 | 0 | **0.900** | **1.000** | 0.944 |

`systematicFailures: []` on both. Identical figures across runs. For context, #19's comment records `recall_user` precision at **0.667–0.750** historically — this is above that band, not below it, so the added rubric did not degrade the memory tool. Stated as observed, not claimed as an improvement: two runs at 3 trials is not enough to establish a shift.

`CORE_PROMPT` 1994 → 2170 tokens against `MAX_SYSTEM_PROMPT_TOKENS` 5000.

## Two harness defects found while measuring, both fixed here

**1. 15 RPM is per-project and the pacing budget was stale.** Two runs aborted on `429 GenerateRequestsPerMinutePerProjectPerModel-FreeTier`. `TRIAL_PACING_MS`'s comment budgets "2 calls / 12s = 10 RPM", but that predates #81's safety de-escalation ladder, which can spend further attempts inside one turn — and the quota is per *project*, so a harness run competes with live Pi traffic.

`TRIAL_PACING_MS` is now **raise-only** overridable via `ROKABOT_TRIAL_PACING_MS`. The human's Gate-1 amendment of 2026-07-29 ("do not tune below 12000") is now **enforced in code rather than stated in a comment** — lowering is refused, not clamped silently. Mutation-probed: removing the floor fails exactly the refusal test. At 30s pacing, zero 429s across three full runs.

**2. Raising the pacing then tripped the hardcoded 900s `testTimeout`**, killing a run at 27/36 turns having produced **no verdict at all**. That config's own comment warns against precisely this ("never become the binding constraint, unlike PR #38's 5000ms default which masked its p95 result with an opaque timeout") — raising pacing recreated the trap. The timeout now derives from the pacing.

## Scope

Prompt and tool-description changes plus harness. No Discord surface, no `searchWeb.ts`, no `search.ts`. Tavily spend is unaffected: at the measured 264 turns/month, even a 100% fire rate is 528 credits against a 1000 free-tier limit.

## Gates

`npx vitest run` **845 / 65** · `npx tsc --noEmit` clean · `biome check .` clean (exit code, not `tail`) · `prettier --check .` clean apart from the git-excluded `docs/.kickoff-next-session.md` · `npm run test:live` **green twice consecutively**.


AlaskanTuna:
@coderabbitai review

PR #95, opened 2026-08-18
fix(discord): stop splitting emoji in half when hard-cutting a long reply

AlaskanTuna:
The swept-vs-sampled follow-up from #92's review. Applying that lesson to `splitResponse` found a real defect on the first sweep.

## The Bug

When `splitResponse` falls through to a hard cut — no newline or space past the halfway mark — it cuts at an arbitrary UTF-16 code unit. Emoji are surrogate pairs, so the cut can land between the halves:

```
input : "🌸🌸🌸🌸🌸🌸🌸🌸🌸🌸"
chunks: "🌸🌸🌸\ud83c"  |  "\udf38🌸🌸🌸"  |  "🌸🌸🌸"
```

Those lone surrogates render as replacement characters. Roka's replies are dense with emoji and kaomoji, so this is her content, not a synthetic edge case.

The fix steps the cut back off a trailing high surrogate, never past index 1 — going further would emit an empty chunk and stall the loop.

## Reachability, Stated Honestly

**Low today.** It needs a reply longer than `discord.maxMessageLength` (1500) with no newline or space at index ≥ 750 in that window, and Roka's replies target a 50–70 word band. The realistic trigger is a long unbroken run — dense emoji, or scripts that do not use spaces.

So: a real correctness defect in a documented contract, cheaply fixed, not a live incident. Worth fixing because the cost is two lines and the failure is user-visible mojibake.

## How It Was Found — the Method Is the Point

`splitResponse` already had eight tests. Every one sampled a fixed length and a fixed shape, and the chunk-length invariant was asserted for exactly one input. None of them could find this.

Sweeping instead — 6 shapes × 121 lengths × 4 limits — surfaced it immediately. Same lesson as #92's unbudgeted newline: **a single worst-case value tests one point on a boundary, not the boundary.**

Three swept invariants added, each stated as a property rather than an example:

- no chunk exceeds the limit, at any length or shape
- no empty chunk for non-empty input (an empty chunk is both an invalid Discord message and a non-terminating loop)
- no chunk contains half a surrogate pair

Plus one concrete regression case, because a sweep that fails prints an offender list while a concrete case prints the actual mojibake.

## Mutation Probe

Reverting the step-back fails both, with the corruption visible in the concrete one:

```
AssertionError: expected [ 'emojiRun max=3 n=2', …(287) ] to deeply equal []
AssertionError: expected [ '🌸🌸🌸�', '�🌸🌸🌸', '🌸🌸🌸' ] to deeply equal [ '🌸🌸🌸', '🌸🌸🌸', '🌸🌸🌸', '🌸' ]
```

Source restored from a byte-verified backup (md5 `e6fd7a99330d67c68f9df4ed1ac48797`).

The existing eight tests stayed green throughout — which is the measurement of what sampled coverage was worth here.

## Not Included

The `maxMessageLength` bound is the other single-value-pinned budget, but sweeping it means touching `messageBuilder.test.ts`, which #92 modifies. Deferred to avoid a conflict; worth doing once #92 lands.

## Gates

- `npx vitest run` — 849 passed / 65 files, exit 0
- `biome check .` — 183 files, exit 0


AlaskanTuna:
@coderabbitai review

PR #97, opened 2026-08-18
docs: drop citation numbering and refresh the README's stale scoping claims

AlaskanTuna:
Both operator requests from the live `/ask` screenshot, plus what a full sweep of `docs/README.md` turned up.

## 1. Citation Numbering Removed

`fitCitations` no longer prefixes `1. 2. 3.`. Live output now:

```
Ara~ it is an open-source agent framework, apparently! (◕‿◕✿)
-# 🌸 searched the wider world • <t:…:R>
-# 🔗 [wavect.io](<…>)  ·  [eigent.ai](<…>)  ·  [theregister.com](<…>)
```

The three tests asserting the numbered format moved with it — deliberately and visibly. The numbering was never load-bearing for the budget arithmetic: the `  ·  ` separator is unchanged and the row only got shorter, so every budget sweep still passes untouched.

## 2. `search_web` → `/ask` in the Tool Footer Table

Requested, and it created a contradiction worth resolving rather than papering over: that column has meant *"a dedicated slash command that surfaces this tool"*, and `/ask` is the general chat surface where **every** tool can fire. Putting `/ask` on one row while the others show `—` is only coherent if the column's meaning is stated.

So the sentence above the table now says it:

> Every tool below is available implicitly in chat — Roka decides when to call it. The **Slash Command** column names the command built around that tool: `/anime` and `/remind` invoke theirs directly, and `/ask` is listed against `search_web` because looking things up is what that command is for. The rest are reached only by talking to her.

That makes the table true rather than minimal. The alternative — `/ask` on all twelve rows — would have made the column identical to "Implicit In Chat" and told the reader nothing.

## 3. What the Sweep Found

Asked to check "some of the sections", with Tool Footer named only as an example. Two further staleness items, neither requested:

**The Tool Footer section never mentioned citations at all.** #92 added the row; the docs still described only the footer. Now covered, including the footer-says-what / citations-say-where split.

**Memory is documented as guild-scoped, and has not been since Phase 27.** Both handlers derive the tenant as `guildId ?? \`dm:${channelId}\`` (`interactionCreate.ts:69`, `messageCreate.ts:149`), so a DM or group chat is its own tenant. Four places still said otherwise, including the **Privacy** section — *"Claims are isolated per guild"* — which is the one that actually matters, since it is what an operator would rely on when disclosing what the bot retains. All four now describe the real boundary.

Verified against the code rather than inferred, and the scan used a pattern confirmed against a known-present line first, per the zero-result-grep rule.

## Not Changed — Flagging Instead

`docs/README.md` line ~318 records the `recall_user` trigger gate as failing, dated 2026-07-29, and points at #39. Fresher measurements exist post-#93, and the recorded 0.667–0.750 precision band no longer matches what is being measured. I left it alone: it is a dated historical statement rather than a false one, resolving #39 is not mine to declare, and confirming it would need live calls on a bucket that is currently in use. Worth a look by whoever owns that gate.

## Mutation Probe

Reinstating the numbering fails exactly the three tests that pin the format:

```
AssertionError: expected '-# 🔗 1. [crunchyroll.com](<…' to be '-# 🔗 [crunchyroll.com](<…'
```

Source restored from a byte-verified backup (md5 `5b4d493092c0b1be61c562cdf4dde039`).

## Gates

- `npx vitest run` — 858 passed / 66 files, exit 0
- `biome check .` — 185 files, exit 0


AlaskanTuna:
@coderabbitai review

PR #101, opened 2026-08-18
feat(discord): accept up to three images on /ask and nudge on files she cannot open

AlaskanTuna:
Both operator requests. No prompt or tool-description changes, so nothing here touches the live gate.

## 1. `/ask` Takes Three Images

Discord has no multi-attachment option type, so `/ask` now exposes one option per slot — `image`, `image2`, `image3` — with the first keeping its original name so existing usage still resolves.

The count is **not** written twice. `MAX_IMAGE_ATTACHMENTS` moved out of `messageCreate.ts` into a new `src/discord/attachments.ts` alongside the type allowlist and the slot-naming function, and the command builds its options in a loop over that ceiling while the handler reads them back through the same naming function. Raising the limit is a one-line change in one file.

The 4 MB per-image cap needed nothing: `MAX_IMAGE_SIZE_BYTES` in `roka.ts` already applies to every `ImageAttachment`, so both surfaces were always sharing it.

`messageCreate.ts` now imports the shared policy, and its three copies of the inline type predicate collapse into `isSupportedImage` — one spelling of one rule.

## 2. An In-Character Nudge Instead of Silence

An unopenable file was filtered out with no signal, so she answered "what's in this?" as though nothing had been sent — indistinguishable from hallucination. She now answers the question as before and adds one canned line:

> Ah — I couldn't open that file, sorry~ Could you tell me what's in it?

Four variants live in `src/discord/responses.ts` beside the other in-character pools. Nothing in `src/agent/prompts/*` moved.

**Applied to both surfaces.** The same silence existed on the mention path, and #19 requires slash and implicit behaviour stay consistent — fixing only `/ask` would have shipped the constraint violated.

**One case deliberately left alone:** a mention carrying *only* an unopenable file and no text still returns early and stays silent. Making her speak there would mean newly responding to something she previously ignored, which is a spam decision rather than a silence-is-misleading fix. The nudge only appears where she was already replying.

## Mutation Probes

Four guards, each probed separately, each failing exactly one test:

| Mutation | Test that fails |
| --- | --- |
| `/ask` reads only the first slot | `forwards every supported image slot, not just the first` |
| `/ask` drops the nudge | `nudges in character when an attachment cannot be opened` |
| mention path drops the nudge | `nudges in character when a mentioned message carries a file she cannot open` |
| command exposes one slot | `offers one attachment slot per image the mention path would accept` |

Sources restored byte-identical afterwards.

## Two Defects Found in My Own Work

Both would have shipped silently:

**An undefined-safety regression.** The old code read one attachment and used `attachment?.contentType`, which tolerates `undefined`. My rewrite filtered on `!== null`, which does not — and `getAttachment` is mocked across four suites as returning `undefined`. The handler's own `try/catch` swallowed the resulting `TypeError` into an error reply, so the first symptom was an unrelated-looking assertion failure three frames away. Now filtered on truthiness.

**A count that silently produced `NaN` under test.** `message.attachments` is a discord.js `Collection` in production but a plain array in the test double, so `.size` arithmetic evaluated to `NaN`, `NaN > 0` was false, and the nudge simply never fired — a test that would have passed by accident forever. The count now materialises through `.values()`, which both types support.

## Gates

- `npx vitest run` — 865 passed / 66 files, exit 0
- `biome check .` — 186 files, exit 0


AlaskanTuna:
@coderabbitai review

PR #103, opened 2026-08-18
docs: describe the quote style Biome actually enforces

AlaskanTuna:
One line in the Code Style bullet. Docs only.

## Why

The bullet said *"single quotes"* flat. Read literally that is wrong for any string containing an apostrophe, and a reviewer on #102 filed exactly that finding against four such strings.

Verified against the config rather than described from memory: `biome.json` sets `javascript.formatter.quoteStyle: "single"`, which in Biome means *prefer* single and switch to double rather than escape a quote inside the string. So the flat reading contradicts the tool the gate actually runs — converting those strings to single quotes produces:

```
× Formatter would have printed the following content:
  - describe('reading what the sender\'s own message shows', () => {
  + describe("reading what the sender's own message shows", () => {
```

and `biome check .` exits 1.

The bullet now says what the formatter does, with the failure mode named, so the next reader — human or automated — does not file it again or "fix" working code into a failing gate.

## Scope

Nothing behavioural. No rule loosened, no gate changed, no permission altered — this corrects a description of tooling behaviour that is independently checkable by running `biome check`.

## Gates

- `npx vitest run` — 871 passed / 66 files, exit 0
- `biome check .` — 186 files, exit 0


AlaskanTuna:
@coderabbitai review

PR #105, opened 2026-08-18
feat(agent): put the finding first when Roka has looked something up

### LINKED ISSUE #94: enhancement(agent): a looked-up answer now competes with the 50-70 word band
Carved out of **#19 item 2** (dialogue delivery optimization), which is the only one of that issue's seven items still open. #19 auto-closed when PR #93 merged — GitHub parsed `Closes #19 items 4, 5, 6 and 7` as a plain `Closes #19`. Items 1, 3, 4, 5, 6 and 7 are genuinely done; this is the remainder, re-scoped against what changed today.

## Why the original framing no longer fits

#19 item 2 asked for three things: in-character synthesis quality, source attribution inside the reply, and length behaviour under the word band. **Two of the three have moved.**

- **Source attribution is largely solved.** PR #91 and #92 render a separate citation row (`-# 🔗 1. [domain](<url>) · …`) on both the slash and mention paths, complementing the tool-usage footer rather than duplicating it. Attribution no longer needs to live inside the prose.
- **The length tension got materially worse, not better.** Two changes compounded it:
  - PR #89 moved `include_answer` from `'basic'` to `'advanced'`, taking Tavily's synthesised answer from roughly 150-250 characters to a measured **67-1157 characters**.
  - PR #93's rubric took `search_web` from a measured **1.7% fire rate** (4 of 236 production turns) to firing on most factual questions.

So Roka now receives far more source material, far more often, while `src/agent/prompts/core.ts` still instructs *"Keep responses between 50-70 words"*. The Phase 15 voice check already measured a tool-call turn at **102 words** against that band, back when the tool fired rarely and returned a fraction of the text.

## What This Issue Is Actually About

Whether the 50-70 word band is still the right instruction for a turn that has just looked something up, and if not, what replaces it. The `## Answering Questions` section already carves out a different balance for direct questions (50-70% answer, 30-50% personality); a searched turn may want its own carve-out, or the band may want to widen conditionally.

## Constraints

- **This moves the system prompt**, so it costs **two consecutive green `npm run test:live` runs** to adopt (~120 paid calls, and note the 15 RPM ceiling is per-project — see the pacing notes on `TRIAL_PACING_MS`). Agree a live-call budget before starting.
- The live gate scores tool *firing*, not reply length, so it will not detect a regression here. **A length or voice measurement needs its own instrument**, and building one is probably the larger half of this task.
- `docs/decisions.md` 2026-07-30 (#52): a prompt addition can regress the tool gate through **recall**, not only precision. Any edit near `## Answering Questions` or `## How You Speak` risks the `search_web` and `recall_user` triggers that currently measure 1.000 and 1.000/0.900.
- Do not widen into the citation rendering — that shipped and is pinned.

Per this repo's standing rule, confirm against source rather than this text: read the word-band instruction in `src/agent/prompts/core.ts` under `## How You Speak`, the `## Answering Questions` carve-out directly beneath it, and the current `search_web` rubric under `## Using Tools`. Re-measure the answer-length range rather than trusting the 67-1157 figure — it was sampled across a handful of queries, not characterised.

AlaskanTuna:
Closes #94 once the live gate passes. **Draft deliberately — see Blocker.**

#94 asks whether the 50-70 word band is still right for a turn that has just looked something up. Measured answer: **the band was not the problem, the padding was.** Baseline replies averaged 87.8 words — 25% over the ceiling — while carrying only 41.0 words of actual finding. This change averages 61.0 words, inside the band, carrying 45.8.

## Measured, Five Live Rounds

Four fixed queries through `generateResponse` inside `withSearchCitations`, real Tavily key.

| Variant | Change | Avg total | Avg finding | Avg closing | Roleplay intro | Single-paragraph |
| ------- | ------ | --------- | ----------- | ----------- | -------------- | ---------------- |
| Baseline | — | 87.8w | 41.0w | 32.8w (+14.0w intro) | 4/4 | 0/4 |
| Attempt 1 | one bullet in Answering Questions | 88.5w | — | — | 4/4 | 0/4 |
| A2 | subsection + Using Tools exception | 90.5w | 44.8w | 41.5w | 1/4 | 0/4 |
| A2b | tightened closing, no list lead-in | 90.8w | 44.5w | 46.3w | 0/4 | 0/4 |
| A2c | closing capped by an 80/20 proportion | 72.3w | 35.5w | 36.8w | 0/4 | 0/4 |
| A2d | proportion reverted, capped by paragraph | 76.3w | 43.3w | 33.0w | 0/4 | 0/4 |
| **A2e** | **+ speech quotas scoped off searched turns** | **61.0w** | **45.8w** | **15.3w** | **0/4** | **2/4** |

## Why Attempt 1 Failed

It added a counter-rule and left the rule it contradicted standing. `## Using Tools` opened with *"weave them into your personality — don't just dump raw data. React to the results in character"* — an instruction to build the exact wrapper the issue asks to remove. Rescoping that one line took the roleplay intro from 4/4 to 1/4 immediately.

Two further contradictions were found the same way and are why the new subsection states its own precedence: `## Answering Questions` allocates 30-50% of the reply to personality, and `## How You Speak` sets a 50-70 word band a searched answer already overran.

## Why the Closing Block Needed a Different Layer

It survived five separate instructions — *never end with a separate block of roleplay*, *one sentence, never a paragraph*, *only if you actually have something to add*, *20% at most*, and *never give your own words a paragraph of their own*. Measured closing size across those attempts: 41.5, 46.3, 36.8, 33.0 words. It was not shrinking because **nothing was ordering it and something else was requiring it.**

`SPEECH_PROMPT` sets quotas — 2-5 bold phrases, kaomoji on 30-40% of responses, soft teasing phrases — that cannot be satisfied inside a factual paragraph. The model manufactured a second paragraph to hold them. Scoping those two quotas off searched turns cut the closing to 15.3 words and produced the first single-paragraph replies of the exercise.

## Rejected, With Evidence

**A proportion is the wrong instrument.** "The finding is at least 80% of the reply" (A2c) shortened both sides and left the ratio intact: total 76 → 72w, but finding fell 43 → 36w and Q1's answer halved from 62w to 28w. Reverted in favour of a paragraph-structure rule, which is countable.

## Not Read as a Regression

Q4's finding mass across the six runs went 34 / 50 / 21 / 54 / 35 / 18 words with no monotone relationship to any change — *"has any anime been delayed this month?"* returns a genuinely varying result set. Called out rather than averaged away.

## Tests

Three pins in `tests/harness/__tests__/tokens.test.ts`, one per load-bearing rule, one assertion each. Each was mutation-probed individually: removing its rule failed exactly that test and left the other two green. Sources restored from filesystem snapshots and confirmed byte-identical by md5.

## Blocker — Do Not Merge Yet

A system-prompt change costs **two consecutive green `npm run test:live` runs** to adopt, per #94's own constraints, because the gate can regress through recall and not only precision. **Neither run has passed.**

The attempt aborted on the harness's own reliability guard:

```
Rate-limit/reliability guard: case "F2" trial 1 returned outcome=fallback
(kind=transient_http) instead of a genuine model turn — aborting the run
rather than scoring a corrupted observation.
```

Underlying cause is `GenerateRequestsPerDayPerProjectPerModel-FreeTier`, `quotaValue: 500` — the harness project's **daily** quota is spent. It resets at midnight Pacific. This is my own fault: I misread a `ps` pattern as "the run died", started a duplicate, and ran two gates concurrently for several minutes.

**Production is unaffected.** `GEMINI_API_KEY` and `GRAPHIFY_GEMINI_API_KEY` are in different projects — verified by a single `generateContent` call on the production key returning HTTP 200 while the harness project was exhausted. The bot on the Pi has its own 500.

Gates that did run: `npm test` **874 / 66** exit 0, `tsc --noEmit` exit 0, `biome check .` exit 0.

## Summary by CodeRabbit

- **Improvements**
  - Lookup and web-search responses now lead with the most relevant findings before additional commentary.
  - Retrieved facts are presented more clearly and confidently, with improved formatting for names and numbers.
  - Search-based responses avoid unnecessary roleplay, preambles, and decorative speech elements.

- **Tests**
  - Added coverage to verify response ordering, search behavior, and formatting rules.


AlaskanTuna:
Marked ready for review to get CodeRabbit's pass while the harness quota is exhausted, so any fix lands before the two adoption runs rather than invalidating them. **Ready for review is not ready to merge here** — the blocker in the description stands until two consecutive green `npm run test:live` runs exist.

AlaskanTuna:
@coderabbitai full review

AlaskanTuna:
## Update — CodeRabbit's Finding Was Real, and the Fix Changed the Result

**Fixed:** the searched-turn rules said they applied *"whenever a tool fetched the answer for you"*. Only **1 of 12** registered tools is `search_web`. The rules were also governing `roll_dice`, `flip_coin`, `search_anime`, `get_anime_schedule`, `get_weather`, `recall_user` and the three reminder tools — and on the dice and coin tools the roleplay reaction is the entire product, so stripping it is a character regression. All four measurement queries were searches, so the instrument structurally could not detect this. Now scoped to `search_web` by name, with explicit precedence for everything else.

**The fix changed behaviour far more than expected, in the wrong direction first.** Naming the tool made the condition unambiguous, and the rules fired at full strength:

| Variant | Avg total | Avg finding | Single-paragraph | Character present |
| ------- | --------- | ----------- | ---------------- | ----------------- |
| A2e (broad wording) | 61.0w | 45.8w | 2/4 | yes |
| A2f (scoped to `search_web`) | 37.3w | ~37w | 4/4 | **no** |
| **A2g (rebalanced)** | **66.0w** | **60.8w** | **3/4** | **yes** |

A2f produced four replies with no voice at all — no closing beat, no kaomoji, three of four *under* the 50-70 band. That breaches `## Hard Boundaries`' "Stay in character at all times. You are Roka, not an AI assistant" in the same prompt. Fixed by making the closing sentence **mandatory rather than permitted** and adding "never drop them entirely — you are still Roka on these turns, not a search engine", and by retargeting the speech-layer rule at the *mechanism* (being pulled into a second paragraph) instead of banning the quotas outright.

**Final position against baseline: total length down 25% (87.8w → 66.0w) while the finding itself is up 48% (41.0w → 60.8w).** The padding was not merely trimmed, it was replaced with information. Average sits inside the stated 50-70 band.

Generalisable, and logged: **narrowing a prompt rule's scope made it stronger, not weaker.** An ambiguous condition fires weakly; a precise one fires at full strength and carries any over-reach with it.

**Also fixed:** the nitpick on `tokens.test.ts`. Pins now cover the scope restriction, the bolding guidance and the closing-block prohibition. Six pins total, each mutation-probed. Note that the quota exemption and the bolding guidance share a source line, so line-deletion fails both — each was re-probed by mutating its own clause, which fails exactly one test.

## Harness Change Included

`tests/harness/env.ts` hardcoded `GRAPHIFY_GEMINI_API_KEY` and read it from the `.env` file rather than the environment, so no override could reach it. It now honours `ROKABOT_HARNESS_KEY` naming which `.env` key funds a run, defaulting to the same key so nothing changes for anyone else. The fail-closed guard is untouched — `ROKABOT_HARNESS_KEY=GEMINI_API_KEY` still throws `Harness resolved the production GEMINI_API_KEY`.

This exists because each key sits in its own Google project with its own daily quota. Exhausting one no longer stalls the gate for a day; you name another.

## Live Gate

Run 1 **green**, on the `DEV_GEMINI_API_KEY` project:

```
truePositives: 18, falsePositives: 0, trueNegatives: 18, falseNegatives: 0
accuracy: 1, precision: 1, recall: 1, systematicFailures: []
```

The recorded pre-change figures were TP=18 **FP=2** TN=16 FN=0, precision 0.900 — the two false positives are absent here. **Not claimed as caused by this change** on one run; run 2 is in progress and will be reported either way.

Gates: `vitest run` 877/66 exit 0, `tsc --noEmit` exit 0, `biome check .` exit 0 — binaries run directly, not through npx or the RTK wrapper, both of which returned misleading exit codes during this work.

AlaskanTuna:
## Live Gate — Two Consecutive Green Runs, Adoption Criteria Met

Both on the `DEV_GEMINI_API_KEY` project, 30s pacing, 36 turns each.

| Run | TP | FP | TN | FN | Accuracy | Precision | Recall | Systematic Failures |
| --- | -- | -- | -- | -- | -------- | --------- | ------ | ------------------- |
| 1 | 18 | 0 | 18 | 0 | 1.000 | 1.000 | 1.000 | none |
| 2 | 18 | 2 | 16 | 0 | 0.944 | 0.900 | 1.000 | none |
| *recorded pre-change baseline* | *18* | *2* | *16* | *0* | *0.944* | *0.900* | *1.000* | *none* |

Criterion is `precision >= 0.80 && recall >= 0.80`. Both runs clear it.

**Run 1's clean matrix was variance, not an effect of this change.** Run 2 reproduces the pre-change baseline exactly, so the two false positives are a property of the fixture rather than something this PR removed. The claim staked in the previous comment — that FP=0 was "not claimed as caused by this change" pending run 2 — resolves to *no causal effect*. Recorded that way rather than quietly keeping the flattering number.

The gate exists to catch regression, including through recall, which #94 called out as the specific risk of editing near `## Answering Questions` and `## How You Speak`. Recall is 1.000 in both runs and there are no systematic failures. **This change is neutral on tool triggering**, which is the result being sought.

PR #106, opened 2026-08-18
feat(discord): read a message forwarded directly to her

### LINKED ISSUE #104: feat(discord): a directly-forwarded message is invisible unless it arrives as a reply
Deferred deliberately from #102, which closed the own-message reading gap for embeds, containers, polls and stickers. This is the one source left on that path.

## The Gap

`message.messageSnapshots` — Discord's payload for a **forwarded** message — is never read on the triggering message. The replied-to path handles it: `messageCreate.ts` walks `referencedMessage.messageSnapshots` and pulls the forwarded content, container text, embed title/description, and images.

So the asymmetry #102 fixed still holds for exactly one element:

| Action | Result |
| --- | --- |
| Reply to a forwarded message and mention her | ✅ she reads the forwarded content |
| Forward a message, mention her in the same message | ❌ she sees only the mention |

The second is the more natural gesture — forward something and ask "what do you make of this?" — and it is currently the one that fails.

## Why It Was Not Done in #102

Stated so the next person does not have to rediscover it:

- The snapshot block is **nested and stateful**. Unlike embeds, polls and stickers, it does not just produce text — it also pushes into `imageAttachments`, so it cannot be lifted into a pure describer the way `describeEmbed` and `describePoll` were.
- `messageCreate.ts` is **mutation-pinned**, carrying assertions from #77 and #29. #102 deliberately added only sources that needed no gate restructuring and left the replied-to path byte-identical. Folding in a stateful nested block would have widened that change past what could be verified the same way.

Neither reason is an argument against doing it — only against bundling it.

## Shape of the Work

The forwarded-snapshot handling in the replied-to branch is the reference implementation; the own-message path needs the equivalent. Watch for:

- **The shared image budget.** Snapshot images must draw on the same `MAX_IMAGE_ATTACHMENTS` ceiling as attachments and embed images, not add a fourth source with its own allowance. Ordering matters: what the sender just handed her should fill slots before a forwarded or replied-to message does.
- **Nested forwards.** A forwarded message may itself carry embeds and containers; the existing block reads snapshot embeds as title + description only, which is narrower than `describeEmbed`. Decide deliberately whether to unify those rather than inheriting the narrower form by accident.
- **Test doubles lie here.** #101 and #102 both lost time to `Message` doubles missing fields that discord.js always supplies — `embeds`, `poll`, `stickers`, a real `toJSON()` on components. A missing field does not throw a useful error; it produces silent green or a wrong answer three frames away. Extend the double to match the platform rather than making production defensive.

Confirm against source rather than this text: `grep -n "messageSnapshots" src/discord/events/messageCreate.ts` should show reads on `referencedMessage` only. #99's coverage table was wrong in two of three columns precisely because it was trusted rather than re-derived, so re-derive this one.

AlaskanTuna:
Closes #104.

`message.messageSnapshots` — Discord's payload for a **forwarded** message — was read only on the replied-to message, never on the triggering one. Forwarding a post and asking "what do you make of this?" in the same breath reached her as a bare mention; replying to that same forward worked. The second is the more natural gesture and it was the one that failed.

## Premise Re-Derived

Per the issue's own instruction, confirmed against source before planning rather than trusted:

```
$ grep -n "messageSnapshots" src/discord/events/messageCreate.ts
237:      if (referencedMessage.messageSnapshots.size > 0) {
238:        for (const snapshot of referencedMessage.messageSnapshots.values()) {
```

Both reads on `referencedMessage`, none on `message`. Premise holds.

## What Changed

The replied-to branch's inline snapshot block is now a shared `describeForwardedSnapshots(snapshots, imageSlots)` returning `{ parts, images }` instead of mutating `imageAttachments` in place — the same extraction shape as `describeEmbed`/`describePoll` in #102. Both paths call it, so a forward reads identically whether it was sent to her or replied to.

## Three Decisions Worth Naming

**Snapshot embeds now go through `describeEmbed`.** The old inline block read them as `title + description` only. A forwarded link keeps its substance in the embed, and the narrow form dropped the author, fields and footer that say where it came from — the exact defect class #102 fixed for own messages. This changes the replied-to path's output too, deliberately.

**Image precedence is explicit:** own attachments → own embed images → own forwarded images → replied-to sources. A forward is someone else's post, so it draws on the same `MAX_IMAGE_ATTACHMENTS` ceiling and only after what the sender attached themselves. Pinned by a test with 2 own + 2 forwarded images asserting the exact resulting triple.

**Trigger scanning is unchanged.** Forwarded text is deliberately *not* added to `triggerScanText`; forwarding a post that happens to contain "roka" should not summon her.

## Verification

- 8 new tests, each mutation-pinned; every probe's failure set matched its prediction:
  - drop own forwarded text → 4 text tests fail
  - drop own forwarded images → 3 image tests fail
  - revert snapshot embeds to title+description → only the link-preview test fails
  - own forward ignores slots already taken → only the precedence test fails
  - drop reply-path forwarded text → only the reply test fails
- Full gate green with exit codes checked on the binaries directly: `tsc --noEmit` 0, `biome check .` 0, `prettier --check .` 0, `vitest run` 0 — **879 tests, 66 files**.

## Two Things Found Along the Way

**The doubles were missing `messageSnapshots`** — fourth time in a row a `Message` double omitted a field discord.js always supplies. Dropping the unreachable `if (!snapshots)` guard from production turned that into 27 loud failures instead of silent green; both doubles were then extended to match the platform.

**One of my new tests was vacuous and only the mutation probe caught it.** The wordless-forward test asserted `.not.toContain('pinged you without saying anything')`, but `replaceUserMentions` strips `<@\d+>` only — with the file's conventional `<@bot-1>` id the mention survived, `content` was never empty, and the bare-ping fallback it pins could never fire. It passed under a mutation that should have broken it. Now uses a numeric id and an exact-equality assertion.

## Not Changed

- Snapshot images are pulled even when replying to the bot, outside the `if (!isReplyToBot)` guard that skips its expression thumbnails. Pre-existing, and unreachable in practice since the bot does not forward.
- When the ceiling is already full, the `(forwarded image(s))` marker is suppressed along with the images, so she is not told about images she cannot see. Pre-existing on the replied-to path and preserved rather than quietly changed.

## Self-Review Notes

Checked while waiting on a review slot; neither is a regression, both are inherited:

- **Forwarded text keeps raw `<@123…>` mentions.** `replaceUserMentions` runs on the triggering message only — it resolves ids through `message.mentions`, which a snapshot does not carry. The replied-to path has always had this; the own path now matches it rather than diverging. Fixing it properly means resolving ids without a mentions collection, which is its own change.
- **Sub-element order differs between the two paths.** Own: container → embeds → poll → stickers → forwarded. Replied-to: content → embeds → poll → forwarded → container → stickers. Cosmetic, and the replied-to order predates this PR; unifying it would move pinned output for no behavioural gain.

Also verified rather than assumed: `Message.js:478` does `this.messageSnapshots ??= new Collection()`, so the field is always present and the non-optional signature is correct — the doubles were the wrong side, not production. `Embed#fields` returns `this.data.fields ?? []`, so routing snapshot embeds through `describeEmbed` cannot throw on a fieldless embed.


AlaskanTuna:
@coderabbitai full review

PR #108, opened 2026-08-18
fix(discord): name a forwarded image she has no room to see

### LINKED ISSUE #107: fix(discord): a forwarded image she has no room for is not mentioned at all
Preserved deliberately in #106 rather than changed, because #106's job was the own-message forwarded path and this predates it on the replied-to path. Splitting it out so it isn't inherited silently.

## The Behaviour

In `describeForwardedSnapshots` (`src/discord/events/messageCreate.ts`), the `(forwarded image(s))` marker is pushed only when the slice actually yielded images:

```ts
const fwdImages = fwdAttachments
  .filter(isSupportedImage)
  .map((a) => ({ url: a.url, contentType: a.contentType! }))
  .slice(0, imageSlots - images.length)
images.push(...fwdImages)
if (fwdImages.length > 0) fwdParts.push('(forwarded image(s))')
```

`imageSlots` is what remains of `MAX_IMAGE_ATTACHMENTS` after the sender's own attachments and embed images. When it reaches 0 the slice is empty, so both the images **and** the sentence saying they exist disappear together.

So a sender who attaches three screenshots and forwards a post containing a photo gets a prompt in which the forwarded photo is not mentioned at all. Ask "what's in the second one?" and she has no good answer — not "I can only see three", but nothing to work from.

## Why It Reads as a Defect

The suppression is backwards relative to what the marker is for. It is not a caption for an image she can see — the image is already in the vision slots and needs no announcement. Its only value is telling her something exists that she *cannot* see. It is dropped in exactly the case where it carries information and kept in the case where it is redundant.

The same file already takes the opposite position one branch over, unconditionally, for the replied-to message's own attachments:

```ts
if (referencedMessage.attachments.size > 0) refParts.push('(attached image(s))')
```

`src/discord/responses.ts` also has `getRandomUnsupportedAttachment()`, added in #101 for the neighbouring case: a file she cannot open is named rather than filtered out silently, because silence there reads as hallucination rather than as a limitation. This is the same failure mode against a different cause — a ceiling instead of a MIME type.

## Shape of the Work

Small, but there is a real choice in it:

- Compute the marker from the **candidate** images (post-`isSupportedImage`, pre-`slice`) rather than the taken ones.
- Decide whether "there were more images than she could see" is worth distinguishing from "there was a forwarded image" — a count is more useful to her than a bare noun and costs a few tokens.
- Whether the replied-to attachment path should be brought onto the same rule, since `(attached image(s))` currently fires even when nothing was taken. Making both consistent is probably the point of doing this at all.

## Verifying

`describeForwardedSnapshots` is unit-testable with no live call; #106 added the fixtures. The pin is a message with `MAX_IMAGE_ATTACHMENTS` own attachments plus a forward carrying an image, asserting the marker survives. Mutation-check it by reverting the marker condition to the taken-images count — it must fail that test and nothing else.

AlaskanTuna:
Closes #107.

The `(forwarded image(s))` marker was keyed to the images actually taken, so when the shared 3-slot ceiling was already full the marker vanished along with the images it was there to announce. A sender who attaches three screenshots and forwards a post containing a photo produced a prompt with no trace of the photo at all — "what's in the second one?" had nothing behind it, and she answered as though the forward carried no picture.

## The Change

Count candidates, not the ones taken, and say how many were withheld:

```ts
const fwdCandidates = fwdAttachments.filter(isSupportedImage).map(...)
const fwdImages = fwdCandidates.slice(0, imageSlots - images.length)
images.push(...fwdImages)

const unseen = fwdCandidates.length - fwdImages.length
if (fwdCandidates.length > 0) {
  fwdParts.push(unseen > 0 ? `(forwarded image(s), ${unseen} not shown)` : '(forwarded image(s))')
}
```

The marker is not a caption for an image she can already see — the image is in the vision slots and needs no announcement. Its only value is naming one she *cannot*, which is precisely where the old form deleted it.

## Two Decisions

**The count, not just the presence.** "There is a forwarded image" leaves her unable to distinguish one she is looking at from one she is missing. `2 not shown` lets her say what she cannot see instead of guessing at it, which is the failure this whole thread started from.

**Unsupported files are not counted.** A forwarded PDF is not an image she is missing; it is a file nothing could have shown her, and #101's reply-side nudge already covers that case. Counting it would invent an image that never existed — the same defect in the opposite direction. Pinned by its own test.

The all-fit case is byte-identical to before, so nothing that reads `(forwarded image(s))` today changes.

## Verification

4 new tests, each mutation-pinned, every probe matching its prediction:

| Mutation | Fails |
| --- | --- |
| Marker keyed to taken images again (the defect) | names-it-anyway, says-how-many |
| `unseen` forced to 0 | says-how-many |
| Candidates stop filtering on supported type | unopenable-file |
| Marker always qualified, even when nothing withheld | does-not-qualify |

The fourth probe exists because the all-fit test survived the first three untouched, and an assertion no probe can break is one I have no evidence about — that is how yesterday's vacuous test in this same file got through to a mutation run.

Gate green with exit codes checked on the binaries directly: `tsc --noEmit` 0, `biome check .` 0, `prettier --check .` 0, `vitest run` 0 — **889 tests, 66 files**.

## Left Alone, Deliberately

One branch over, the replied-to path pushes `(attached image(s))` whenever `attachments.size > 0` — unfiltered, so replying to a message carrying only a PDF tells her there is an image when there is none. That is the same class of defect as this one, pointing the other way: this PR was about a marker that vanished when it mattered; that one appears when it shouldn't. It has no filed issue and no reported symptom, and folding it in would move replied-to output under a PR about forwarded images. Happy to file it separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of forwarded images when attachment limits are reached.
  * Added clear context indicating how many supported forwarded images were omitted.
  * Prevented unsupported forwarded files from being counted as missing images.
  * Ensured fully included forwarded images are not incorrectly marked as unavailable.


AlaskanTuna:
@coderabbitai review

Issue #100, opened 2026-08-18
feat(agent): let Roka see video, hear audio, and read documents within limits the Pi can survive


AlaskanTuna:
Operator request, 2026-08-18: *"I've been thinking of wanting Roka be able to 'see' and 'hear' videos, audios with reasonable limits (otherwise it'd nuke my host machine), as well as able to read documents, but I don't know what's the best and free way to engineer this."*

**This is a research-and-design issue first.** It should produce a costed recommendation before any implementation, because the binding constraints are all outside this repo — the model's modality support, the free tier's token accounting, and a Raspberry Pi 5 with a container memory cap.

## Today's Baseline

`ALLOWED_IMAGE_TYPES` is `image/png`, `image/jpeg`, `image/gif`, `image/webp` — declared identically in `src/discord/events/interactionCreate.ts` and `src/discord/events/messageCreate.ts`. Up to `MAX_IMAGE_ATTACHMENTS` (3) per turn, each capped at 4 MB (`src/agent/roka.ts`), downloaded, `sharp`-processed to JPEG, and attached as `inlineData`.

**Everything else is silently dropped** — no error, no acknowledgement. That silent-drop behaviour is being fixed separately alongside raising `/ask` to three attachments; this issue is about actually *supporting* the other modalities.

## Research Questions — Answer These Before Designing

1. **Does the configured model accept video and audio at all?** `gemini.model` is `gemini-3.5-flash-lite`. Lite variants have historically carried reduced modality support. **Verify against Google's own model page, not a blog** — and if Lite cannot, the honest options are a per-modality model switch (with its own cost) or declining the feature, not silently degrading.
2. **Inline data vs the Files API.** Inline has a request-size ceiling; the Files API handles larger payloads and is free, but adds an upload round trip, a storage lifetime, and a second failure mode on the hot path. Which applies at what size?
3. **What does this cost in tokens?** Video is sampled per second and audio is billed per second, so a short clip can dwarf an entire text conversation. The free tier limits **tokens per minute** as well as requests — a single video turn could trip TPM where 15 RPM was never the issue. Measure before promising.
4. **Do documents even need new plumbing?** Gemini accepts PDFs directly. If so, document support may be far cheaper than video or audio and worth splitting out and shipping first.

## The Host Constraint Is the Real Design Problem

The operator's *"otherwise it'd nuke my host machine"* is the governing requirement, and it is not primarily about Gemini:

[...]

PR #110, opened 2026-08-18
docs: cost multimodal intake against the Pi's measured limits

AlaskanTuna:
Research and design for #100. **No implementation** — this ends at a costed recommendation for Gate 1, as the issue asks.

Adds `docs/multimodal.md`. Deliberately not linked from `docs/README.md`: that file is operator-facing reference for behaviour that exists, and this describes behaviour that doesn't yet.

## Verdict

| Modality | Recommendation | Cap | Blocked On |
| --- | --- | --- | --- |
| PDF / documents | **Ship now** | 10 MB | Nothing — rides the existing path |
| Audio | **Ship second** | 8 MB | Global in-flight byte budget |
| Video, low media resolution | **Ship last** | 10 MB, ~60 s | Global budget + streaming size guard |
| Video, default resolution | **Decline** | — | 3× tokens for no benefit at this size |

Memory and latency are unrelated constraints and both land on 10 MB. That agreement is the main reason to trust the number.

## Three Premise Corrections

- **The cap is 512 MB, not 8 GB.** `docker-compose.yml:13` is `mem_limit: 512m`. The issue reads as though the Pi's 8 GB were the budget; the bot gets one sixteenth of it.
- **`ALLOWED_IMAGE_TYPES` is not duplicated any more.** The issue lists de-duplication as an open constraint; #101 already consolidated it into `src/discord/attachments.ts`.
- **Flash-Lite is not modality-reduced.** The issue suspected Lite variants might not take video/audio. Google's model page: input is "Text, Image, Video, Audio, and PDF". No model switch needed, nothing blocked at the model level.

## The Findings That Changed the Design

**The container is SIGKILLed with no log output.** Measured, both swap configurations: `exit 137`, stdout and stderr empty. No catchable error, no final log line — under `restart: unless-stopped` the bot just vanishes and returns. The only evidence is `docker inspect`. A cap you can't observe approaching has to be designed against with margin, because there's no runtime signal to react to.

**Docker silently granted 512 MB of SD-card swap.** `MemSwap=1073741824` — defaulted to 2× `mem_limit` because `memswap_limit` was never set. So the container doesn't fail fast at 512 MB; it thrashes the SD card first (observed: RSS pinned at ~547 MB while allocation climbed past 990 MB), blowing `turnDeadlineMs` and wearing the card, then dies anyway. **Recommending `memswap_limit: 512m`** — one line, and the only config change this research proposes.

**An attachment costs ~4.7× its size in RSS.** Measured across 4/10/20/40/60 MB in a real 512 MB cgroup. It's 4.7× because the payload exists four times over — `arrayBuffer`, the `Buffer.from` copy, the base64 string, and the JSON body — summing to 4.66×. Measurement and arithmetic agree, so it extrapolates.

**V8's heap limit doesn't help.** Node 24 *is* cgroup-aware (259 MB heap limit inside 512 MB, verified rather than assumed), but Buffers are external to the heap. Two of the 4.7 multiples never count against it and never throw a catchable OOM — they go straight to container RSS, where the kernel is the first thing that notices.

**The binding constraint is concurrency, and there is no global cap.** `src/discord/concurrency.ts` is one active request *per channel* with no global limit, so K busy channels means K simultaneous footprints. Against the Pi's measured 55.59 MiB idle RSS, total in-flight budget is ~97 MB of payload at zero margin. **A global in-flight byte budget is the prerequisite for audio and video** — not just a per-turn size cap.

**`buffer-then-check` can't bound a missing Content-Length.** `downloadImage` reads the header, then `await response.arrayBuffer()`, then checks the real size — the middle step buffers the whole body first. Safe today because Discord's CDN always sends the header, and harmless at 4 MB regardless. Not safe at video sizes; video needs a streaming counter that aborts mid-transfer.

## Where I'd Push Back on the Issue

It asks for explicit **duration** caps. Duration isn't knowable without decoding the file, and decoding costs the same memory the cap exists to protect. The byte cap is the enforceable proxy; the duration figures in the doc are what a byte cap *implies* at typical bitrates, not independently enforced limits. Enforcing real durations means `ffprobe` in the image plus a decode pass, and I don't recommend paying that for a bound bytes already achieve. Flagged rather than silently substituted.

## One Number I Could Not Get

Free-tier **TPM**. Google moved per-tier limits off the public rate-limits page into AI Studio, which needs an authenticated session. The 250,000 figure in the doc is from a developer-forum thread, isn't Google's own documentation, and doesn't name Flash-Lite. It's labelled as unverified where it appears.

It only matters for video — audio and PDF are an order of magnitude below RPM 15 either way. **Gate 1 action: read it off [AI Studio's rate-limit page](https://aistudio.google.com/rate-limit)**, ten seconds of operator time. Video shouldn't start before that's confirmed.

## Method Note

The first latency measurement said **85 KB/s**, which would have forced me to recommend declining audio *and* video outright. That was `httpbin.org` throttling, not the Pi's uplink — Cloudflare's upload endpoint gives 2.5 MB/s, and a `generate_204` cross-check was discarded because it answered after 64 KB and never sampled throughput at all. Recorded in the doc rather than buried, because a wrong number there would have produced a confident wrong recommendation on the whole feature.

## Verification

Docs-only, no source touched: `prettier --check` 0, headings and table headers follow the TitleCase rule in AGENTS.md. Every figure is either measured here with the command shown, or cited to Google's own model documentation, with the single exception called out above. Zero Gemini calls. All Pi access read-only — `docker stats`, `docker inspect`, `free`, `df`, and one 5 MB throughput POST; nothing restarted, redeployed, or written.


AlaskanTuna:
@coderabbitai review

Issue #111, opened 2026-08-18
chore(agent): pinClaim/unpinClaim have no production callers, so the 20-claim eviction exemption config.yml documents is unreachable


AlaskanTuna:
## Found While Scoping #25

`pinClaim` and `unpinClaim` (`src/agent/memory/memoryClaims.ts:301,305`) have **zero production callers** — grep confirms every call site is a test. `pinned` is `0` on all 286 active claims in production, verified via read-only SSH.

`pinned` is load-bearing in three places:

- **First** ordering key on the recall tool path (`pinned DESC, salience DESC, last_seen_at DESC`)
- A `+1` scoring term on the passive path (`src/agent/memory/retriever.ts`)
- The **only** exemption from `evictOverflow` (`memoryClaims.ts:170`) — `config.yml`'s `maxActiveClaimsPerUser: 20` comment says "pinned claims are exempt"

So the exemption `config.yml` documents cannot currently be triggered by anything. Nothing decides which facts are important enough to protect from the 20-claim ceiling; eviction is purely salience/recency.

## Not In Scope Here

Kept separate from #25 Phase 1 deliberately — that phase changes *ranking* of what's shown, not *retention* of what's kept, and the two shouldn't be conflated in one diff.

## Open Questions for Whoever Picks This Up

- Should something call `pinClaim` automatically (e.g. `sourceKind: 'explicit'`, or a salience threshold), or is this meant to be a future user-facing action (a command, a reaction)?
- Does #25 Phase 2's `search_memories` change what's worth pinning?

Low priority, no labels, not implemented.

PR #112, opened 2026-08-18
feat(agent): rank recall_user's facts by relevance instead of recency alone

### LINKED ISSUE #25: feat(agent): semantic memory retrieval via embeddings + guild-wide memory search

AlaskanTuna:
Closes #25 Phase 1. Phase 2 (`search_memories`, embeddings) is not in this PR.

## Premise, Verified Against Production

Read-only SSH to `rpi5-ubuntu`, not assumed. 286 active claims, 24 (guild, subject) pairs. **11 of 24 (46%) sit at exactly the 20-claim storage ceiling** with `pinned = 0` on all 286 rows and 60% of salience floor-bunched at 0.25 — so for those 11 subjects, `recall_user`'s static `pinned DESC, salience DESC, last_seen_at DESC` ordering collapses to plain recency, and 5 of each subject's 20 stored facts are permanently outside the 15-fact recall window.

## The Issue's Proposed Fix Could Not Have Worked As Written

#25 proposed ranking by cosine similarity to the current message. `recall_user`'s declared schema is `user_name` only; the rest of its context arrives via `toolContext.state`, which carried `_systemPrompt`, `participants`, `_userId`, `_channelId`, `_guildId` — **and no message text**. It was structurally incapable of relevance ranking, embeddings or not, until the message reaches it.

There are also two retrieval paths, and only one had the problem. `src/agent/memory/retriever.ts`'s passive path already runs FTS5 `MATCH` against the current message with a scoring function (salience × source weight, confidence, recency, pinned, an FTS-hit bonus, a routed-predicate bonus). `recall_user` called a *different*, unscored `getActiveClaims`. This PR threads the message into `recall_user` and has it reuse the retriever's existing scorer, rather than building parallel ranking logic.

## Changes

- **`src/agent/roka.ts`** — `stateDelta` (both the real shape and the test-injectable `TestTurnRequest`) carries `_userMessage: userMessage`, already in scope where the delta is built.
- **`src/agent/memory/retriever.ts`** — new exported `retrieveForSubject(guildId, subjectUserId, message, limit)`. Same scoring as `retrieveForTurn`, extracted a shared `routedPredicatesFor` helper rather than duplicating it; no speaker anchoring, cross-user expansion, or token budget, since `recall_user` targets one named person by design. Filters `needsReview` claims, matching the passive path — this **excludes** them from `recall_user` for the first time (previously it read the un-filtered `getActiveClaims`); a claim needing review shouldn't surface either way, and it directly serves the operator's guild-scoping ruling below.
- **`src/agent/tools/recallUser.ts`** — `RecallUserParams` gains `message: string`; calls `retrieveForSubject` instead of the static ordering.
- **`src/agent/tools/index.ts`** — `execute` reads `toolContext.state.get('_userMessage') ?? ''` and threads it through both call sites (named recall and self-recall).

## On Guild Scoping (Operator's Ruling)

*"Should be applied to everyone and restricted within per-server — no cross-server, no DM/group chat retrieval."* Unaffected by this change: `retrieveForSubject` takes `guildId` from the same `toolContext.state._guildId` the tool already fails closed on, and its query is `WHERE guild_id = ? AND subject_user_id = ?` — no change to isolation, and the existing DM-tenant test (`interactionCreate.memory.test.ts`) still passes unmodified in substance (only its mock needed `retrieveForSubject` added via `importOriginal`, since the file mocks the whole module).

## Tests

`tests/harness/__tests__/recallRelevance.eval.test.ts` (new) — same production-modelled seed as the plan (20 active claims, no eviction, one fact buried by age). No model calls: `recall_user` is pure over SQLite, pass/fail is exact string containment.

- **BEFORE ×3** — regression guard. An empty/irrelevant message still misses the buried fact; proves the fix is the message, not a general loosening of the cap.
- **AFTER ×3** — the case's query surfaces the fact.
- **FTS-isolated case** — `misc` predicate, whose only keyword is `'misc'`, so `routeTopics` grants no bonus to either side. Added because a first pass at the other three cases turned out to be discriminated entirely by the routed-predicate *category* bonus, not literal FTS token matching — a mutation that zeroed `searchClaimIds`'s output was silently absorbed with no failures. This case fails when it's zeroed and nothing else does.
- **needs_review exclusion** — a claim flagged `needsReview: true` is excluded even when the message matches its value exactly.
- **`memoryTools.test.ts`** — a `FunctionTool`-level test (`recallUserTool.runAsync`) proving `toolContext.state`'s `_userMessage` actually reaches ranking end-to-end, not just the internal function. Added because an early probe changing the tool-layer wiring passed cleanly — the fixture had been calling `recallUser()` directly and never exercised `tools/index.ts` at all.

Four mutation probes (message dropped at the tool layer, `needsReview` filter removed, FTS zeroed, each independently) — each broke exactly the test built to catch it and nothing else. Sources restored from filesystem snapshots, confirmed byte-identical by md5.

Existing `memoryTools.test.ts` cases needed only `message` added to already-passing calls; ordering assertions among same-predicate/same-source claims held unchanged, because with no relevance signal, `scoreClaim` degenerates to a function of recency alone — same order as the code it replaced.

## Filed, Not In This PR

**#111** — `pinClaim`/`unpinClaim` have zero production callers, so the 20-claim eviction exemption `config.yml` documents is currently unreachable by anything. Found while reading the eviction path; kept separate since it's about retention, not ranking.

## Gates

`vitest run` 900/67 exit 0 (was 896/67 before four coverage-gap fixes) · `tsc --noEmit` exit 0 · `biome check .` exit 0 · `prettier --check .` clean except the pre-existing, git-excluded `docs/.kickoff-next-session.md`. Binaries run directly.


AlaskanTuna:
@coderabbitai full review

PR #114, opened 2026-08-18
feat(discord): name a shared GIF or video instead of dropping it

AlaskanTuna:
Follow-up to #100. Ships the part that is genuinely self-contained; the rest is written up as a comment on #100 rather than built.

## The Gap

A pasted Tenor or Giphy link arrives as a `gifv` embed whose **only** content is `embed.video` — no title, no description, no fields. `describeEmbed` returns `null` when those are all empty, and nothing anywhere read `embed.video`. So a message whose entire substance was a shared GIF reached her as a bare mention.

```ts
if (embed.video) parts.push(embed.data.type === 'gifv' ? 'animated GIF' : 'video')
```

## Why This Names It Instead of Showing It

The obvious version of this change — adding `embed.video?.url` to the existing `embed.image?.url ?? embed.thumbnail?.url` fallback so the GIF reaches the vision slots — **would ship a defect**, and I verified the mechanism rather than reasoning about it:

1. `embed.video.url` on a gifv serves an **MP4**.
2. Everything bound for the vision slots goes through `processImageForGemini`. Confirmed against sharp directly: an MP4 buffer raises `Input buffer contains unsupported image format`.
3. `imageProcessor.ts:58` catches that and returns **the original undecoded buffer labelled `image/jpeg`**.
4. `messageCreate.ts:263` independently hardcodes `contentType: 'image/png'` for embed images.

So the net effect would be MP4 bytes sent to Gemini declared as a JPEG — a burned vision slot, wasted tokens, and a `logger.warn` as the only trace. **There is a test pinning that this cannot happen**, and mutating the image chain to include `embed.video?.url` fails exactly that test and nothing else.

Actually showing her the motion needs #100's video intake — the caps, the global byte budget, the streaming guard. This PR deliberately does not build a second pipeline alongside it.

## One Correction Worth Recording

discord.js's `video` getter surfaces only `url`, `proxyURL`, `height`, `width`. Discord's payload also carries `content_type` and a `description` (alt text), but the getter drops both — they are reachable only through `embed.data.video`. Relevant to whoever implements the intake: `content_type` would let it route by declared type instead of guessing, and it is *there*, just not on the typed accessor.

## Verification

4 tests, 3 mutation probes, each matching prediction:

| Mutation | Fails |
| --- | --- |
| Drop the `embed.video` read | names-a-GIF, calls-it-video, still-frame |
| Stop distinguishing gifv from video | names-a-GIF, still-frame |
| Route the video URL into the vision slots | never-sends-video-to-slots |

The third is the one worth having long-term — it is a regression guard against the plausible-looking "improvement" described above.

`tsc` 0, `biome` 0, `prettier` 0, `vitest` 0 — **904 tests, 67 files**. No live calls.

## Deliberately Not Here

Two things need their own design and are written up on #100 instead: the `messageUpdate` timing problem (link-preview embeds are frequently absent on the initial `messageCreate`, and this bot registers no `messageUpdate` listener at all), and `/ask`'s complete lack of URL detection. Both raise a behavioural question — does she answer twice, or wait? — that is the operator's call at Gate 1, not mine to settle in a field read.


AlaskanTuna:
@coderabbitai review

PR #116, opened 2026-08-18
feat(discord): simplify /stats Who She Knows Best to the 5 most recent members' latest memory

AlaskanTuna:
Operator request: replace "Who She Knows Best" — currently ranked by claim count with a decay-weighted salience pick for the quote — with something simpler: up to 5 most recent guild members, each showing their single latest memory.

## What Changed

`MEMORY_DETAIL_SQL.topRememberedMembers` replaces a three-CTE count+decay-salience query with one CTE: rank each member's claims by `last_seen_at DESC`, keep rank 1 per member, order the surviving members by that same recency, cap at 5.

Dropped along with the old ranking:
- **Salience-based decay formula** (`salience + 0.5 / (1 + ageDays/30)`) — recency alone decides now.
- **Claim count** — no longer computed, no longer rendered (`RememberedMember.count` removed from the interface and the card).
- **`nowMs` parameter** on `topRememberedMembers()` — only existed to feed the decay formula; the function is now a pure query over stored timestamps, no wall-clock dependency.

## One Behaviour Change Worth Stating, Not Leaving Implicit

The old query used a `LEFT JOIN`, so a member whose only claims needed review still appeared in the list with a null predicate/value, rendered as a generic fallback quote ("I remember a little something about them~"). The new query filters `needs_review = 0` before ranking, so that member has no presentable row and **doesn't appear in the top 5 at all** — a real member is shown only when there's a genuine memory to show for them.

This directly serves the "simple and reliable" framing: the old LEFT JOIN existed specifically to handle this edge case with a fallback string; the new design has no edge case to handle, because a member with nothing presentable is simply absent.

## Tests

- Rewrote the two tests that encoded the old salience/count contract.
- Added a dedicated recency-vs-salience test — and mutation-probing it caught a real weakness in its first draft: the two predicates I originally chose (`hobby`/`likes`) happened to also sort in the "correct" order alphabetically, so the test kept passing under a mutation that deleted the recency ordering entirely and left only an alphabetical tiebreak. Rewrote using `likes`/`pets`, where alphabetical order actively disagrees with recency order, so the test can only pass if `last_seen_at` is genuinely driving the pick.
- Added an explicit "caps at 5 when more are eligible" test — the user's stated "(max)" is otherwise only implied by the SQL `LIMIT`.
- Updated the pre-existing 2026-07-24 human-directed pin (the cross-query loop asserting `needs_review = 0` literally appears in this query's SQL) — comment updated to note the 2026-08-19 simplification; the pin itself needed no change and still holds.

Three mutation probes (raise the LIMIT, drop the needs_review filter, remove recency from the tiebreak) each fail exactly the test(s) built to catch them. The needs_review probe also independently tripped the pre-existing 2026-07-24 pin, which is a second layer catching the same regression.

## Gates

`vitest run` 949/68 exit 0, `tsc --noEmit` exit 0, `biome check .` exit 0. No live model calls.

## Summary by CodeRabbit

* **New Features**
  * Memory highlights now show each member’s most recent eligible memory.
  * Results are limited to the five most recently updated members.
  * Entries requiring review or without eligible memories are excluded.

* **UI Updates**
  * Memory displays now show the member name and quoted memory without a numeric count.
  * Improved fallback handling for unavailable or departed members.


AlaskanTuna:
@coderabbitai review

AlaskanTuna:
Both findings verified against the code rather than applied reflexively. **Declining both**, with evidence.

## 1. "Future-dated provenance comments" — Declined, timezone artifact

The review reasons in UTC; this repo dates in local time.

```
$ date       → Wed Aug 19 01:45:01 +08 2026
$ date -u    → Tue Aug 18 17:45:01 UTC 2026
$ git log -1 --format='%ad'  → Wed Aug 19 01:37:58 2026 +0800
```

Commits carry `+0800`, and the existing logs already use local dates — `docs/progress.md` has a `## [19/08/26]` heading and `docs/decisions.md` has `2026-08-19` entries, both written earlier today in this timezone. The comment is same-day, not future-dated. Changing it to 2026-08-18 would make it disagree with the commit that introduces it and with every neighbouring log entry.

## 2. "Exclude empty fields before ranking" — Declined, unreachable

Correct that `NOT NULL` permits `''`. It cannot occur, on either column, and the guard would be coding around a scenario the write path structurally prevents.

**There is exactly one production INSERT into `memory_claim`** — `src/agent/memory/memoryClaims.ts:231`, inside `assertClaimInTransaction`. Every other `INSERT INTO memory_claim*` in `src/` targets the `memory_claim_fts` mirror and is trigger-driven from rows that already passed it.

**`value`** — that function calls `assertSafeValue(op.value)` before the insert, which throws on `!isSafeFactScalar(...)`, whose first condition is:

```ts
if (!s.trim() || s.length > maxLen || /[\r\n]/.test(s)) return false
```

An empty or whitespace-only value throws `Claim value is unsafe` and never reaches SQL.

**`predicate`** — stored as `normalizePredicate(op.predicate)`, whose return type is `PredicateId` and whose final line is `return SYNONYMS[normalized] ?? 'misc'`. It resolves to a known predicate or the literal `'misc'`; there is no path returning `''`.

This also matches a precedent set on #96 in this repo: defensive de-duplication was deliberately *not* added there because the `toolsUsed` Set already guaranteed the property, and `AGENTS.md`'s coding guidelines say plainly — *no error handling for impossible scenarios*. Adding `predicate <> '' AND value <> ''` plus a regression test for a state the schema's writer cannot produce would be speculative work, and the requested test would be unable to construct its own precondition through the public API.

If the concern is that a *future* write path might bypass `assertClaim`, the durable fix is a `CHECK (length(trim(value)) > 0)` constraint at the schema level rather than a filter in one read query — but that's a separate change with a migration, not this PR.

PR #117, opened 2026-08-18
fix(discord): stop claiming a replied-to file is an image she can see

### LINKED ISSUE #109: fix(discord): replying to a message with no image still tells her one is attached
The mirror image of #107, on the same marker family and the same path. #107 fixed a marker that vanished when it carried information; this is one that appears when there is nothing to announce.

## The Behaviour

`src/discord/events/messageCreate.ts`, in the replied-to branch:

```ts
if (referencedMessage.attachments.size > 0) refParts.push('(attached image(s))')
```

The condition is attachment *count*, with no type filter. Every other place the codebase turns an attachment into something Roka is told about runs it through `isSupportedImage` first — the own-message vision slots, the forwarded-snapshot images, and `/ask`'s option handling all do. This line does not.

So replying to a message whose only attachment is a PDF, a zip, a text file or a video puts `(attached image(s))` in her prompt, and there is no image. Nothing else in the turn contradicts it.

## Why It Matters

It is the failure mode the operator described that started this thread — half-baked answers about things she cannot actually see — except here she is not merely uninformed, she is misinformed. Asked "what's in the image?" about a reply carrying a PDF, the prompt says an image is attached, so the plausible completion is a description of an image that does not exist. #101 added `getRandomUnsupportedAttachment()` for the neighbouring case precisely because silence read as hallucination; asserting a nonexistent image is worse than silence.

The count is also unqualified in the other direction: five attachments and three attachments produce the same string, so she cannot tell how many she is missing. #107 settled that question for the forwarded marker (`(forwarded image(s), 2 not shown)`); this line predates that decision.

## Shape of the Work

Small, and #107 established the pattern to copy:

- Filter through `isSupportedImage` before deciding whether to push the marker.
- Decide what a reply carrying only unopenable files should say — nothing at all is defensible, but so is naming it, since she can see the message has *something* attached and saying nothing invites the same guessing. Note that the replied-to path has no equivalent of the mention path's unsupported-attachment nudge, because the nudge is about what the sender just handed her.
- Bring the count onto #107's rule so a reply with more images than the ceiling admits says how many are missing, rather than the current unqualified plural.

## Verifying

Unit-testable with no live call; the fixtures exist in `src/discord/__tests__/messageCreate.test.ts` from #106 and #107. The pin is a reply whose only attachment is a PDF, asserting the prompt does not claim an image. Mutation-check by removing the type filter — it must fail that test and nothing else.

AlaskanTuna:
Closes #109. The mirror of #107, on the same marker family and the same file — #107 fixed a marker that vanished when it carried information; this is one that appeared when there was nothing to announce.

## The Defect

```ts
if (referencedMessage.attachments.size > 0) refParts.push('(attached image(s))')
```

Attachment *count*, no type filter. Every other place this codebase turns an attachment into something Roka is told about runs it through `isSupportedImage` first — the own-message slots, the forwarded-snapshot images, `/ask`'s options. This line did not.

So replying to a message whose only attachment is a PDF put `(attached image(s))` in her prompt with no image behind it. She is not merely uninformed there, she is **misinformed**, and the plausible completion is a description of an image that does not exist. #101 added the unsupported-file nudge precisely because silence read as hallucination; asserting a nonexistent image is worse than silence.

## The Fix

Counts supported-image *candidates* rather than raw attachments, and qualifies the shortfall the way #107 settled for the forwarded marker:

- **Unopenable file only** → no marker at all. A PDF is not an image she is missing.
- **Image fits** → `(attached image(s))`, byte-identical to today.
- **Images do not fit** → `(attached image(s), 2 not shown)`.

`refImagesTaken` is now computed once and reused by the extraction block, rather than the filter+slice being written twice with the marker keyed off a third, different expression.

## One Behaviour Change Beyond the Type Filter, Stated Rather Than Slipped In

Replying to **Roka's own message** carrying an image previously produced a flat `(attached image(s))` while image extraction was deliberately skipped (expression thumbnails waste tokens). That is the same false assertion in different dress — she was told she could see something she was never given.

It now reports `(attached image(s), 1 not shown)`. The deliberate skip is unchanged; only the claim about it is corrected. Called out because it is a real change the issue did not explicitly request, and it falls out of applying one coherent rule — *the marker describes what she was actually given* — rather than special-casing three conditions.

## Tests

Four, in a new `naming a replied-to image she cannot see` block, mirroring #107's structure on the reply path.

Four mutation probes:

| Mutation | Fails |
| --- | --- |
| Revert to raw unfiltered `attachments.size` | unopenable-file, not-shown-count, own-thumbnail |
| Drop the `N not shown` qualifier | not-shown-count, own-thumbnail |
| Count images *taken* instead of candidates (#107's defect on this path) | not-shown-count, own-thumbnail |
| Remove the marker entirely | image-fits, not-shown-count, own-thumbnail |

The fourth exists because the common-path guard (`still names a replied-to image when one fits`) survived the first three — correctly, since it is a regression guard for behaviour that must stay byte-identical, but an assertion no probe can break is one I have no evidence about. It breaks under removal, so it is live.

Sources restored from filesystem snapshots between probes, confirmed byte-identical by md5.

Materialised via `[...attachments.values()]` so the count reads the same off a discord.js `Collection` in production and the plain arrays some doubles in this file use — the trap that has bitten this file three times.

## Gates

`vitest run` **953/68** exit 0, `tsc --noEmit` exit 0, `biome check .` exit 0. Binaries run directly. No live model calls.


AlaskanTuna:
@coderabbitai review

Issue #121, opened 2026-08-18
chore(agent): price the 3-image ceiling before changing it


AlaskanTuna:
Operator's question, 2026-08-18: *"whether is it wiser to have Roka to intake 3 attachments at once or strictly just one at a time, I suddenly feel like it's not sustainable to let prod free tier Gemini API key to handle this many as these cost relatively more tokens than normal message queries."*

Separate from #100 because that issue prices modalities we have not shipped; this prices the one already running on every mention and every `/ask`. Research only — no change made.

**Recommendation up front: do not lower the ceiling. The instinct is directionally right but the magnitude is small, nothing image-related binds any limit, and the one number needed to decide it is not being recorded.**

## What an Image Actually Costs

Gemini charges 258 tokens for an image with both dimensions ≤384px; larger images are tiled, at 258 tokens per tile, with the crop unit `floor(min(w,h) / 1.5)`.

Applying that to what `processImageForGemini` emits (max 1024 longest side, min 512 shortest):

| Output size | Tiles | Tokens | |
| --- | --- | --- | --- |
| 384×384 | 1 | **258** | flat rate |
| 512×512 | 4 | 1,032 | sharp's `MIN_DIMENSION` |
| 1024×768 | 4 | 1,032 | typical 4:3 |
| 1024×1024 | 4 | 1,032 | sharp's `MAX_DIMENSION` |
| 2048×1536 | 4 | 1,032 | twice the size, same cost |
| 1024×576 | 6 | 1,548 | 16:9, the worst common shape |

**The non-obvious part: above 384px the cost depends on aspect ratio, not resolution.** A 512×512 image and a 2048×1536 image both cost 1,032 tokens. So `MAX_DIMENSION = 1024` is not costing anything a smaller cap would save, and shrinking images *within* the current range is not a lever at all. The only per-image reduction available is getting both dimensions to ≤384, which is a 4× cut.

## Against a Measured Text Turn

Production `tokens_in_est` over 259 turns: min 3,975, **mean 5,389**, max 7,809 — dominated by the four-layer system prompt.

| Turn shape | Tokens in | vs a text turn |
| --- | --- | --- |
| text only | 5,389 | 1.00× |
| + 1 image | 6,421 | **1.19×** |
| + 2 images | 7,453 | 1.38× |
| + 3 images | 8,485 | **1.57×** |
| + 3 images, and she searches | 16,970 | **3.15×** |
[...]

PR #124, opened 2026-08-18
fix(agent): let an outright request to remember something actually reach remember_user

### LINKED ISSUE #118: fix(agent): remember_user has never fired in production, leaving #111's auto-pin dormant
## Found While Implementing #111

`remember_user` has **never fired in production**. Measured read-only against the Pi, not inferred:

```sql
SELECT COUNT(*) FROM response_events WHERE tools_used LIKE '%remember_user%';  -- 0
SELECT COUNT(*) FROM response_events WHERE tools_used LIKE '%recall_user%';    -- 15
SELECT COUNT(*) FROM response_events;                                          -- 259
```

And no `explicit` claim has ever existed, at any status:

```
legacy/active=176   legacy/rejected=135   legacy/superseded=17
passive/active=111  passive/rejected=19   passive/superseded=81
```

`src/agent/tools/rememberUser.ts:31` is the only writer of `sourceKind: 'explicit'`, so the two facts corroborate each other: the tool is registered, its sibling `recall_user` fires fine, and this one has never once been called across 259 turns.

## Why It Matters

It makes #111's shipped behaviour dormant. The operator chose to auto-pin explicitly-remembered claims so something is finally protected from the `maxActiveClaimsPerUser: 20` ceiling — 11 of 24 production subjects sit pegged at it. That rule is correct and tested, but it can only fire when an `explicit` claim is written, which today never happens.

## Candidate Causes — Not Yet Investigated

1. **The prompt may be routing around it.** `core.ts` says: *"When users newly share personal information, call `remember_user` immediately and silently in the same turn; **skip facts already in What You Remember**."* Passive extraction runs on its own schedule and has produced 111 active claims. If the passive path reliably lands the fact first, that final clause tells her to skip the tool — permanently, and by design.
2. Tool description or parameter shape making it unattractive to the model relative to answering directly.
3. A silent failure path, though the `recall_user` counterpart working argues against a wiring problem.

Cause 1 is the most likely and the most interesting, because it means the tool is not broken so much as **structurally superseded by the passive extractor**, and the prompt line that does it reads as a sensible efficiency instruction.

## Scope

Investigation first, no fix proposed. If cause 1 holds, the question is whether `remember_user` should exist at all, or whether being *told* a fact should mark it differently from having it *inferred* — which is exactly the distinction #111 now depends on.

**Any prompt-layer change inherits the two-green-live-run cost** that #94 and #105 paid, so establish the cause before touching `core.ts`.

### LINKED ISSUE #123: test(agent): the live gate never loads search-web.jsonl, so it has only ever measured recall_user
## The Gap

`tests/harness/tool-trigger/search-web.jsonl` was added by #93 and **nothing has ever loaded it**. It appears in exactly one commit — the one that created it — and no source file references it.

`toolTrigger.live.test.ts:36` calls `loadCaseSet()` with no argument, and `loadCaseSet(path = DEFAULT_CASE_SET_PATH)` defaults to `recall-user.jsonl`. So the live gate runs one fixture: 12 recall_user cases × 3 trials = the 36 turns it reports.

## Why It Matters

**"The live gate passed" has meant `recall_user` only.** #93 measured `search_web` at 18/18 by pointing the harness at the fixture directly, and that measurement has not run since. Every two-green-run adoption since then — #105's searched-turn prompt rework most relevantly, which *specifically* changed how `search_web` results are presented and rewrote the `## Using Tools` section — verified that `recall_user` did not regress and said nothing about `search_web`.

No regression is known or suspected. The point is that the gate could not have told us either way, while its name and its use in adoption decisions imply it would.

## Fix

Load both fixtures and score them separately. `loadCaseSet` already takes a path, `runCaseSet` already keys off `header.tool`, and `scoreCaseSet`/`meetsLiveVerdict` are per-case-set — so this is wiring, not new machinery.

Cost: roughly doubles a gate run, 12 more cases × 3 trials, so ~36 extra turns and ~20 minutes at the 30s pacing floor. Worth knowing before adopting the next prompt change rather than after.

## Related

**#118** would want a third fixture for `remember_user`, and would hit this same wiring. Worth fixing this first so that lands into a gate that already runs more than one case set.

AlaskanTuna:
Closes #118. Also carries the privacy-floor work that #118's measurements exposed, and #121's first item.

## 1. An outright request to remember something now reaches `remember_user` (#118)

`remember_user` had never fired in 259 production turns. #118's own "cause 1" was right, and it is now confirmed: the background extractor (`extractionInterval: 20`, batched, monitored channels) already owns facts mentioned in passing. So the tool is not meant to duplicate it — its job is the one thing the extractor cannot do, which is mark a fact as deliberately entrusted (`sourceKind: 'explicit'`), the only thing #111/#120's auto-pin can protect from the `maxActiveClaimsPerUser: 20` ceiling.

Two changes, measured one at a time:

- **Tool description** now names the *phrasings* of a request rather than categories of fact. Measured before/after on the same fixture: `"I want you to remember…"` and `"…don't forget"` both went from silently doing nothing to firing.
- **`core.ts`** keeps `skip facts already in What You Remember`, but gains an exception for an outright request. Without it a fact passive extraction had already landed could *never* be pinned — precisely the case that matters when 11 of 24 production subjects sit pegged at the ceiling.

Deliberately **not** done: pushing the prompt to fire on inferred facts too. That chases recall the extractor already delivers, at a round-trip per chatty turn.

Also fixes #123 in passing — `search-web.jsonl` shipped with #93 and was never loaded, because the gate called `loadCaseSet()` with no argument. "The live gate passed" has only ever meant `recall_user`, including for #105, which specifically changed `search_web` presentation. The gate now runs one scored `it` per fixture.

## 2. A privacy floor on stored facts

The extractor prompt has always banned storing *"age or birthday, … school or workplace names, … medical/health details"* — but that was prose aimed at one model on one path. `remember_user` had no guard at all, and `normalizePredicate` **degrades an unknown predicate to `misc` rather than rejecting it**, so `birthday` became `misc` and the value was stored verbatim.

New `privacyGuard.ts` blocks only what has no legitimate use in a character bot's memory: contact details, money, government identifiers, credentials. Two signals, because either alone leaks — the key catches `password` → `"hunter2"`, and the value catches an address filed under an innocent name, which is the only signal the passive path still has once its key has been normalized away.

**Deliberately allowed**, and this is the point: `birthday`, `age`, `diet` ("allergic to peanuts"), `general_occupation` ("works night shifts as a nurse"), `nickname`. The controlled vocabulary already encodes "general, not specific" — the job, never the employer; the chosen name, never the legal one — and being asked outright is consent. A mutation probe that adds `birthday` to the denylist fails a test, so "not too strict" is enforced by the suite rather than by judgment.

Calibrated against 32 legitimate facts (0 false positives) and 22 sensitive ones (0 misses). The claim-level throw reuses the exact string `'Claim value is unsafe'` because `extractor.ts` and `memoryMigration.ts` both skip on that message; reword either side and one bad value stops being skipped and aborts an entire background batch. A test pins that coupling.

Bundled in, because it is the same defect class in code this PR already touches: `saveFact`'s rejection was discarded, so a fact the prompt-safety guard dropped still came back as `"Remembered"`. The legacy store feeds the facts envelope (`roka.ts`), so a rejection meant the fact landed **nowhere** while Roka said otherwise.

## 3. Images count against `tokensInEst` (#121, item 1)

`tokensInEst` counted every image as zero, so image cost was invisible in `response_events`. It now adds `GEMINI_IMAGE_TOKENS` per image, and 0 for documents (billed per page, not per image).

The number is measured, not cited, and the measurement **overturned the working assumption**. `countTokens` against `gemini-3.5-flash-lite` returns **1089 tokens for every square image from 64×64 through 1024×1024**, easing only at extreme aspect ratios (1081 at 2:1, 1056 at 4:1), against a text-only baseline of 2. Google's published tile model — 258 per tile, one tile below 384px — does not describe this model's billing.

Retracted along the way: a conclusion that `processImageForGemini`'s upscale to `MIN_DIMENSION = 512` pushed small images into a 4× bracket, and that declining to upscale would cut those turns 4× for free. Both false — cost never depended on size. The first probe used a 1×1 image as its "baseline", so its control was a member of the treatment group and never isolated per-request overhead.

The consequence for #121 stands and is firmer: **the only lever on image cost is how many images a turn carries, never how big they are.** Shrink-don't-cut was never available, which also settles the compression question.

## Verification

`tsc` clean · Biome clean · **1019/1019** unit tests (from 982) · 13 mutation probes, every one observed failing and every file restored to checksum.

Live tool-trigger gate, 3 trials each:

| Fixture | Precision | Recall | |
| --- | --- | --- | --- |
| `remember_user` | 1.000 | 1.000 | 36/36 turns correct |
| `recall_user` | 0.900 | 1.000 | reproduces the historical baseline exactly |
| `search_web` | 1.000 | 1.000 | first execution ever (#123) |

> [!IMPORTANT]
> **This has one green live run, not the two that adopting a prompt change requires.** Both dev API keys hit their 500 RPD ceilings; production's key is not available for this. #94 is exactly why the rule exists — there, run 1's clean matrix turned out to be variance rather than improvement. The second run is deferred to the next quota window and tracked on #118. Merging before then is a deliberate call, not a satisfied gate.


AlaskanTuna:
---

## Update — Validated Against Production, And One Self-Inflicted Regression Fixed

**The privacy floor was audited against live data.** The guard's logic was run in place over the Pi's database (counts and predicate names only — no stored value left the host, none entered the transcript):

```
user_memory:   0 rows, 0 would be refused
memory_claim:  566 rows, 0 would be refused
```

566 real claims, zero false positives. Considerably stronger evidence of "not too strict" than the 32-fact synthetic corpus.

**That audit also caught a regression in this PR.** Noticing `user_memory` is empty in production (claims are canonical under `claimsBackend: true`) prompted a re-read of the honesty fix, which was wrong:

`saveFact` rejects on **key or value**; `assertClaim` checks the **value only**, and lets `normalizePredicate` fold an awkward key into `misc`. The two guards disagree, and the disagreement is load-bearing. My original early-return on `saveFact` failure therefore discarded the claim whenever only the *key* was unusable — most plausibly a `fact_key` over `MAX_FACT_KEY_LEN` (64) from a verbose model. That is precisely the claim this tool exists to write, so the fix defeated #118's purpose in a case #118 cares about.

Now reported on the pair: failure only when **neither** store accepted the fact. A verbose label no longer costs the explicit claim. Pinned by a test asserting the `misc` claim still lands with `sourceKind: 'explicit'`, and by two probes — one reinstating the early return, one restoring the original always-success lie.

Running total: **1020/1020** unit tests, 15 mutation probes.

> [!NOTE]
> CodeRabbit's status check reports `success`, but it posted no review — neither a verdict phrase nor a rate-limit notice. That is the known skip-by-default behaviour on repos under 10 stars, which still sets `state=success`. **Treat this PR as unreviewed by CodeRabbit**, not as having passed it.

AlaskanTuna:
## Gate Met — Two Green Live Runs, All Three Fixtures

The bar this PR was held against is now satisfied. Both dev API keys refreshed, and the confirming full-gate run came back clean:

| Fixture | TP | FP | TN | FN | Precision | Recall |
| --- | --- | --- | --- | --- | --- | --- |
| `recall_user` | 18 | 2 | 16 | 0 | 0.900 | 1.000 |
| `search_web` | 18 | 0 | 18 | 0 | 1.000 | 1.000 |
| `remember_user` | 18 | 0 | 18 | 0 | 1.000 | 1.000 |

Zero `RESOURCE_EXHAUSTED`, zero reliability-guard aborts, 108/108 turns completed.

**Two independent green runs per fixture:**

| Fixture | Run 1 | Run 2 |
| --- | --- | --- |
| `recall_user` | full gate | full gate (confirming) |
| `search_web` | full gate | full gate (confirming) |
| `remember_user` | 3-trial solo run | full gate (confirming) |

`recall_user` reproducing **0.900 / 1.000 exactly** across two independent runs is the reassuring part — run-to-run stability is precisely what this rule exists to check, and it is what #94 failed when its first clean matrix turned out to be variance rather than improvement.

**One precision note, since the runs and the branch are not simultaneous.** Commit `111f60d` landed after the gate had already collected its modules, so the gate exercised the pre-`111f60d` tree. That commit touches only `rememberUser.ts`'s execute path and one test — no prompt layer, no tool description, no fixture. The trigger-determining surface the gate measured is identical to what merges here, and a storage-side change cannot affect whether the model calls the tool. Stated rather than glossed.

Issue #128, opened 2026-08-19
feat(agent): let Roka watch a short video, low resolution only, behind a streaming size guard


AlaskanTuna:
Split out of #100, now closed with its research complete. Steps 5 and 6 of the Suggested Sequencing in `docs/multimodal.md`, filed together because video is the only consumer of the streaming guard.

**Blocked on #126** (global in-flight byte budget). Largest payloads, and the only modality whose token cost can bind before RPM.

## Part 1 — Streaming Size Guard

`downloadAttachment` in `src/agent/roka.ts` reads `content-length`, then buffers the **whole** body, and only then checks the size. That is safe today for two reasons that both stop being true at video sizes: Discord's CDN always sends the header, and 4–10 MB is harmless to buffer regardless.

Replace buffer-then-check with a running byte counter that **aborts mid-transfer** once the cap is exceeded. A missing or lying `content-length` must not be able to exhaust memory before the check runs.

## Part 2 — Video, Low Media Resolution Only

**Decline default resolution.** Two independent reasons, and the second is new since #100 was written:

1. **Token cost.** 300 tokens/second at default, 100 at low — 3× for no benefit at these sizes.
2. **The byte cap stops holding.** TPM is now measured at **250,000 input tokens/minute** (`GenerateContentInputTokensPerModelPerMinute-FreeTier`, scoped to `gemini-3.5-flash-lite`, #125), and it counts *every* input token — system prompt and history included, which production averages at 5,543 per turn. A minute saturated at `rpm: 15` therefore spends ~83,000 on text before any media, leaving ~11,100 per turn: **~110 s of low-resolution video, ~37 s at default.** The recommended 10 MB byte cap buys roughly 40–160 s at typical Discord bitrates. So at low resolution the byte cap already holds video inside TPM with no decode pass; at default resolution it does not.

Cap **10 MB**, set by upload latency against `gemini.timeout`.

## Acceptance Criteria

- Media resolution is pinned to low explicitly, with the request asserted in a test rather than the setting assumed.
- The streaming guard aborts a transfer that exceeds the cap, tested with a response that omits `content-length` and with one that understates it.
- Video bypasses `sharp` — handed MP4 bytes it raises `Input buffer contains unsupported image format`, and its catch relabels them `image/jpeg`, which silently misdeclares the file (this is exactly why #114 named a linked GIF rather than showing it).
- An oversized video is refused in character, never dropped silently.

## Not In Scope

[...]

Issue #127, opened 2026-08-19
feat(agent): let Roka hear an audio clip within the byte budget


AlaskanTuna:
Split out of #100, now closed with its research complete. Step 4 of the Suggested Sequencing in `docs/multimodal.md`.

**Blocked on #126** (global in-flight byte budget). The doc's wording: audio has *"smallest payloads, cheapest tokens, no new failure modes once (3) exists."* It should not ship before that budget does.

## Costed Already — Do Not Re-derive

- **32 tokens per second**, so 1,920 per minute of audio. A five-minute clip is ~9,600 tokens.
- **RPM binds first by an order of magnitude.** Audio never approaches the TPM ceiling, which is now measured rather than assumed: **250,000 input tokens/minute**, `quotaId GenerateContentInputTokensPerModelPerMinute-FreeTier`, scoped to `gemini-3.5-flash-lite` (#125).
- **Byte cap 8 MB**, set by upload latency against `gemini.timeout`, not by container memory.

## What To Build

Admit audio MIME types alongside the existing image and document routing, under the byte budget from #126.

The plumbing already exists and is proven: #119 established that ADK forwards an arbitrary `inlineData` mimeType untouched, verified by a contract test driving a real `Runner` against a capturing model. Audio must bypass `sharp` exactly as documents do — handed non-image bytes, `sharp` throws, and its catch returns the undecoded bytes relabelled `image/jpeg`, which misdeclares the file and makes it unreadable at the other end. `src/agent/roka.ts` already branches on `contentType.startsWith('image/')` for this reason.

## Acceptance Criteria

- Supported audio types are declared in one place and admitted on both the `/ask` and mention paths — the type lists in `interactionCreate.ts` and `messageCreate.ts` have drifted apart before.
- An oversized or unsupported audio file is refused **audibly**, in character, not dropped silently.
- The 8 MB cap is pinned by a matched pair of tests either side of the boundary, as the 4 MB / 10 MB image-versus-document split was in #119 — so both halves of the limit are load-bearing.
- Audio does not pass through `sharp`, asserted directly rather than inferred.

## Not In Scope

Telling Roka *how* to talk about audio she has heard. That is a prompt-layer change and inherits the two-green-live-run cost that #94 and #105 paid; `docs/multimodal.md` says explicitly it should not be bundled here.

PR #129, opened 2026-08-19
feat(discord): bound in-flight attachment bytes across every channel, not just per channel

### LINKED ISSUE #126: feat(discord): bound in-flight attachment bytes across all channels, not just per channel
Split out of #100, which is now closed: its research is complete and `docs/multimodal.md` carries the costed design. This is step 3 of that document's Suggested Sequencing, and the doc's own note is that it is **"the prerequisite for anything larger… worth landing on its own with its own tests, because it is the piece that prevents the OOM."**

## The Problem

`src/discord/concurrency.ts` enforces one in-flight request **per channel**, with no global cap. K busy channels therefore means K simultaneous attachment footprints, and each one is roughly **4.7×** the payload on the heap — `arrayBuffer` + `Buffer` copy + base64 + JSON body. Buffers are external to the V8 heap, so a cgroup-aware heap limit does not catch this; the container simply dies.

Today's per-turn caps are 4 MB per image (×3) and 10 MB per document. Nothing bounds the sum across channels.

## What To Build

A **global in-flight byte budget of 32 MB**, admission-controlled before download begins — not after, which is the same mistake the streaming guard exists to fix (see #128). At worst case that holds residency near 20% of the container's 1 GB `mem_limit`.

Per `docs/multimodal.md`, the per-turn caps do **not** move: 10 MB and 8 MB are set by upload latency against `gemini.timeout`, not by container memory, so raising `mem_limit` left them where they were.

## Acceptance Criteria

- A turn whose attachments would push total in-flight bytes past the budget is refused or queued, and says so in character rather than dropping silently — the silent-drop failure mode is what made #118 and the pre-#114 embed handling hard to diagnose.
- The budget is released on every exit path, including download failure, model error, and timeout. A leak here degrades into a permanent refusal, so this needs a test that drives a failing turn and asserts the counter returns to zero.
- Concurrent turns across different channels are bounded in aggregate — the existing per-channel guard is not sufficient and should not be conflated with this one.
- Tunable in `config.yml` with an inline comment, bounded by `NUMERIC_BOUNDS` in `src/config.ts`, per the project's configuration convention.

## Not In Scope

Duration limits. `docs/multimodal.md` pushes back on that framing explicitly: duration is not knowable without decoding, and decoding costs the very memory the cap protects. The byte cap is the enforceable proxy.

AlaskanTuna:
Closes #126. Step 3 of `docs/multimodal.md`'s sequencing — the prerequisite for audio (#127) and video (#128), and the piece that prevents the OOM.

## The Problem

`src/discord/concurrency.ts` bounds how many turns *one channel* runs. It bounds nothing globally, so K busy channels means K simultaneous attachment downloads. Each costs ~**4.7×** its own size in RSS — `arrayBuffer` + `Buffer` copy + base64 string + JSON body all coexist — and two of those multiples are `Buffer`s, which live **outside the V8 heap** and so never raise a catchable heap error. The container is SIGKILLed with empty stdout and stderr.

## What Landed

`src/discord/byteBudget.ts` — a 32 MB global ceiling, reserved **before** the download starts. A check that runs once the bytes have landed is a measurement, not a guard.

- **Sizing.** Discord states an upload's size up front, so the common path reserves what will actually arrive. An embed image or a link resolved by HEAD states no size and reserves its type's ceiling instead — an unknown has to cost the most it could, or the budget bounds nothing. A stated size *above* the ceiling is clamped to it, because the download refuses on `Content-Length` before buffering, so those bytes are never actually held.
- **Refusal is in character**, via the existing busy reply — not a new path, and never a silent drop. Silent drops are what made #118 and the pre-#114 embed handling so hard to diagnose.
- **Release on every exit path** — success, download failure, model error, timeout — from the same `finally` that releases the per-channel guard. The reservation is taken with nothing between it and that `try`, so no throw can slip between taking the bytes and handing them back.
- `discord.maxInFlightAttachmentBytes` in `config.yml`, bounded in `NUMERIC_BOUNDS`, contract in `docs/trd.md`.

## The Bound That Is Not Obvious

The config `min` is `MAX_ATTACHMENTS × MAX_DOCUMENT_SIZE_BYTES` (30 MB), not 1. Below that, a maximal turn could never be admitted **even on a wholly idle bot** — it would be refused permanently rather than merely delayed. Raising `MAX_ATTACHMENTS` from 3 to 4 without raising the budget would silently create exactly that, so a test asserts the relationship from the constants themselves rather than from a comment.

## One Refactor, and Why It Was Forced

The two per-attachment size caps moved from `roka.ts` into `src/agent/attachmentLimits.ts`. Four test suites replace `roka.js` with a double that exports only `generateResponse`, so a *value* imported from there arrives `undefined` inside them — `reservationFor` would have returned `NaN`, and `NaN > budget` is false, so the guard would have admitted everything while every test still passed. The leaf module removes that failure mode rather than documenting it. `roka.ts` imports them back; behaviour is unchanged.

## Verification

1013 tests pass, `test:perf` passes, tsc / Biome / Prettier clean. 31 new tests, each assertion in its own `it` so mutation pinning means something.

Every new test was pinned by mutation, with the failure count predicted before running — all nine matched:

| --- | --- | --- |
| Drop `release` in the `/ask` finally | 2 | 2 |
| Drop `release` in the mention finally | 2 | 2 |
| `tryReserve` always admits | 8 | 8 |
| Drop the oversize clamp | 2 | 2 |
| Unknown size reserves nothing | 2 | 2 |
| `release` without the zero clamp | 1 | 1 |
| Swap the image/document ceilings | 4 | 4 |
| Drop `clearInterval` on the refusal path | 1 | 1 |

The first mention-path probe returned 1 where the `/ask` one returned 2, which is what surfaced that only the *failing* mention turn was pinned and not the succeeding one. That test — and one asserting the typing indicator stops when a mention turn is refused, so she is not left visibly typing forever — exist because of that probe.

## Not In Scope

Duration limits, per #126 and `docs/multimodal.md`: duration is not knowable without decoding, and decoding costs the very memory the cap protects. The streaming size guard that replaces `buffer-then-check` is #128, where video needs it — today's `Content-Length` check is safe because Discord's CDN always sends the header and harmless at 4–10 MB regardless.


AlaskanTuna:
## Reviewed By Hand — Approving, With One Nit And A Merge-Order Note

Read rather than skimmed, because the green CodeRabbit check on this repo means nothing: it reports `success` while posting no review at all — no verdict phrase, no rate-limit notice — which is its skip-by-default behaviour under 10 stars. You were right to say so rather than let the check stand in for a review.

### What I Checked, And What Held

**Every attachment is actually covered.** This was my main worry: the reservation reads `imageAttachments`, but the mention path also collects embed, forwarded and referenced images. They all push into that *same* array (lines 269, 277, 309, 346, 353) and `reservationFor` runs at 403, after all of them. Covered.

**Release is on every exit path.** `release(reservedBytes)` sits in the same `finally` as `markFree` on both handlers, and both refusal paths return before anything is held — the mention path clearing `typingInterval` first, which is the leak your probe found.

**Unknown sizes cost the ceiling.** Embed images push with `contentType: 'image/png'` and no `size`, so they reserve the full 4 MB. Conservative in the right direction — an unknown that reserved zero would bound nothing.

**The refactor is justified, and I verified the mechanism rather than taking it on faith.** Four suites mock `agent/roka.js` with a factory returning only `{ generateResponse }`. A constant imported from there arrives `undefined` inside them, so `Math.min(size, undefined)` is `NaN`, and `inFlight + NaN > budget` is `false` — the guard admits everything while every test stays green. That is the same failure shape we keep hitting, and moving the constants to a leaf module removes it structurally instead of documenting it. Good catch.

**The config floor is right.** `MAX_ATTACHMENTS × MAX_DOCUMENT_SIZE_BYTES` = 30 MB against a 32 MB default. Deriving it from the constants rather than a comment is what makes raising `MAX_ATTACHMENTS` fail loudly instead of silently creating permanent refusals.

### One Nit, Not Blocking

`markBusy(channelId)` sits between `tryReserve` and the `try` on both handlers, so "nothing between it and the try" is very slightly off. It is a `Set.add` and cannot realistically throw, so the leak is theoretical — but if you ever want it airtight, moving `markBusy` inside the `try` costs nothing, since `markFree` on a channel that was never marked is a no-op `Set.delete`.

### A Consequence Worth Stating Out Loud

At the 32 MB default, one turn carrying three 10 MB documents reserves 30 MB and leaves 2 MB — so it effectively blocks every other channel until it completes. That is the byte budget working as designed rather than a defect, and the busy reply makes it visible. Just worth being explicit that the tuning is deliberately tight, and that the knob to loosen it is `discord.maxInFlightAttachmentBytes`.

### Merge Order

Holding this until #124 lands, not because of anything here. Both touch `roka.ts` around `downloadAttachment` — this PR moves the size constants out, #124 changes the return type and adds `GEMINI_IMAGE_TOKENS` — so they will conflict. #124 is the older branch and closes #118, #121 and #123, so it goes first; this one takes `main` afterwards.

AlaskanTuna:
## Took Main, Took the Nit

Merged `origin/main` (not rebased — this branch is published, and rewriting it is a Critical Do-Not). Pushed as `dfa1623`.

### The Merge Was Clean, Which Is Exactly When to Check

Git auto-merged `roka.ts` without a conflict. That is the case worth distrusting, not trusting, so I verified the result by content rather than by the absence of markers:

- `MAX_IMAGE_SIZE_BYTES` / `MAX_DOCUMENT_SIZE_BYTES` are declared **only** in `src/agent/attachmentLimits.ts`, with `roka.ts` importing them at line 20 and using them at the `limit` branch.
- #124's `downloadAttachment` return type survived intact — `{ data, mimeType, tokens }`, with `GEMINI_IMAGE_TOKENS` for images and `0` for documents.
- `imageTokens` still accumulates into `tokensInEst`.

Both changes coexist; neither silently reverted the other. Gated the merge commit on its own before layering anything on top, so a failure would have been attributable to one or the other: 1051 tests pass, tsc / Biome / Prettier clean.

### The Nit — Taken, Because You Were Right About the Claim

`markBusy(channelId)` did sit between `tryReserve` and the `try`, so "nothing between it and the try" was an overstatement in my PR text. Rather than soften the wording I made the wording true: `markBusy` now runs as the first statement *inside* the `try` on both handlers.

I checked your no-op claim rather than taking it on faith — `markFree` is `activeRequests.delete(channelId)`, and `Set.prototype.delete` on an absent key returns `false` without throwing. So the move costs nothing and the invariant is now structural instead of argued.

Re-ran the mutation probes afterwards, since restructuring a handler can quietly unpin the tests that guard it. All three still fail as predicted: dropping `release` from the `/ask` finally fails 2, from the mention finally fails 2, and `tryReserve` always admitting fails 8.

### The Tightness, Stated Out Loud

Agreed it should not be left for someone to discover. It is now on the record in two places rather than in a PR thread: the `config.yml` inline comment and the `docs/trd.md` contract both say that a single turn of three 10 MB documents reserves 30 MB, that the default therefore sits only 2 MB above its own floor, that this is the budget working rather than a defect, and that `discord.maxInFlightAttachmentBytes` is the knob to loosen it.

### Still Unreviewed by CodeRabbit

Same as before — treat the green check as meaning nothing. Your read is the review this has had.

AlaskanTuna:
## Re-Reviewed At `dfa1623` — Merging

**On the clean auto-merge.** I predicted a `roka.ts` conflict, so its absence was the thing to distrust, and you were right to verify by content rather than accept it. I checked independently and both changes survived:

- Size caps are **imported** at line 20 from `attachmentLimits.js`, with no surviving declaration in `roka.ts` — this PR's refactor intact.
- #124's accounting is whole: the `GEMINI_IMAGE_TOKENS` import, `tokens:` on the image return (695), the `imageTokens` accumulator (859/865/869), and its term in `tokensInEst` (1074).

My conflict prediction was simply wrong — the two changes touched different lines of the same region — and harmlessly so. Gating the merge commit alone before touching anything else was the right sequencing regardless; it is what would have made a failure attributable.

**The nit is properly closed.** `markBusy` is now the first statement inside the `try` on both handlers, with the reason in a comment rather than in a PR description nobody reads later. You fixed the code to match the claim instead of softening the claim, which is the better direction, and you checked the no-op assertion rather than taking it from me — `activeRequests.delete` on an absent key returning `false` without throwing is exactly it.

Re-running the mutation probes after restructuring the handlers is the part I would most easily have skipped. Moving a statement into a `try` is precisely the kind of edit that can unpin the tests guarding it while everything stays green, and 2 / 2 / 8 matching prediction is what makes the restructure trustworthy rather than merely tidy.

**Tightness is on the record** in `config.yml` and the `trd.md` contract rather than in this thread, which is where it needed to be.

**On the two stable `recall_user` false positives** — agreed, and worth keeping. Stable across two independent runs makes them a characterisable behaviour rather than noise, so they are diagnosable if they ever start mattering. Not chasing them now: precision 0.900 clears the bar and they have been steady across every run since #94.

Note for the record, as with #124 and #125: CodeRabbit's green check here means nothing — it posted no review, neither a verdict nor a rate-limit notice. This PR has had two hand reviews instead.

PR #130, opened 2026-08-19
feat(agent): let Roka hear an audio clip within the byte budget

### LINKED ISSUE #127: feat(agent): let Roka hear an audio clip within the byte budget

AlaskanTuna:
Closes #127. Step 4 of `docs/multimodal.md`'s sequencing, on top of the byte budget from #129.

## The Trap This Feature Sits On

Gemini's audio page documents six MIME types: `audio/wav`, `audio/mp3`, `audio/aiff`, `audio/aac`, `audio/ogg`, `audio/flac`. **`audio/mpeg` is not among them** — and `audio/mpeg` is the registered type for an MP3 (RFC 3003), which is what Discord labels the single commonest audio upload there is.

So admitting only the documented list would have refused nearly every mp3 anyone sends, and passing Discord's name straight through would have been a guess about an undocumented alias. Instead: `audio/mpeg` is **admitted** at the Discord boundary and **renamed** to `audio/mp3` in `geminiMimeType`, at the one place bytes reach the model. That is correct whether or not the API also accepts `audio/mpeg`, so it needed no live call to settle.

The inverse also holds and is tested: `audio/mp4` — what an `.m4a` arrives as — is *not* accepted. Starting with `audio/` is not the same question as whether the model can hear it.

## What Landed

- **`ALLOWED_AUDIO_TYPES` in `src/discord/attachments.ts`**, the same single declaration the image and document sets already live in. Both surfaces read it.
- **8 MB ceiling**, and `sizeLimitFor` in `attachmentLimits.ts` is now the one function mapping a type to its ceiling — used by *both* `downloadAttachment` and the in-flight byte budget, so the two cannot disagree about the same file. Before this, audio would have inherited the document ceiling and the budget would have reserved 10 MB for a clip the download refuses at 8.
- **Audio bypasses `sharp`**, which the existing `!isImage` branch already did — but asserted directly rather than inferred, per the AC.
- **`tokensInEst` charges audio 0**, deliberately. Audio is billed per second, and seconds are not knowable without decoding — the same argument the doc makes against enforcing duration caps. Left at zero rather than estimated, and tested.

## Two Decisions Rather Than Inherited Behaviour

**The mention path now admits media, not images only.** The AC requires audio on both surfaces, and a bespoke audio-only predicate there is precisely the drift that let #100's silent-drop survive. Switching the trigger message's filter to `isSupportedMedia` **also admits PDFs on the mention path** — completing the follow-up #119 deferred until `messageCreate.ts` was clear. That is a real behaviour change beyond "audio" and is flagged here rather than left to be discovered. The forwarded, referenced and embed sub-paths stay images-only, because their text markers describe what they carry as images; widening those means rewriting those markers.

**`attachment_url` stays images-only.** `resolveImageUrl` is the path that makes the Pi fetch a host a user named, so its type set is a security decision, not a feature one — the SSRF guard covers the host, never the payload. Its description already says "an image" rather than "a file", so the slot does not over-promise. Widening it deserves its own change with its own tests, not a ride-along inside a modality PR. The reasoning is now a comment on the option itself.

The `attachment_1..3` descriptions **are** updated — they now read "image, PDF or audio clip". A description that under-promises is a feature nobody discovers.

## Verification

1081 tests pass (+30), `test:perf` passes, tsc / Biome / Prettier clean.

Six mutation probes, failure count predicted before each run:

| --- | --- | --- |
| Drop `audio/mpeg` from the allowed set | 4 | 4 |
| `geminiMimeType` stops renaming | 1 | 1 |
| Audio inherits the document ceiling | 3 | 3 |
| `isSupportedMedia` drops audio | 6 | 6 |
| Mention path back to images only | 3 | 3 |
| Route everything through `sharp` | ≥5 | 7 |

The last was bounded rather than pinned because I could not predict how the token assertions would fall out. Its 7 are 2 PDF + 3 audio + 2 token, and the audio pair — `sends an audio clip to the model as audio` and `hands the model the clip exactly as it arrived, and says so` — are among them, which is the AC's "asserted directly rather than inferred".

**The 8 MB cap is pinned by a matched pair at 9 MB**: audio is refused there, a document is accepted there. Either alone would pass with audio inheriting the document limit; together they make the audio ceiling load-bearing.

### One Test Was Reading Stale State

The new mention-path assertions passed in isolation and failed in the suite. The cause was not the code: the sibling `describe` blocks each carry a local `beforeEach` with `vi.clearAllMocks()`, and mine did not, so `mock.calls[0][0]` was reading the first call made anywhere earlier in the file. Worth recording because the first hypothesis — a leak in the new byte budget — was wrong, and instrumenting `inFlightBytes()` across all 66 tests to check showed it never left zero. The budget does not leak; the test was reading the wrong data.

An existing test also had to be repointed: it used a **PDF** as its "file she cannot open" fixture on the mention path, which this change makes openable. Same trap #119 hit. It now uses `application/zip`, with a comment saying why it must not be a PDF or an audio clip.

## Not In Scope

Telling Roka *how* to talk about audio she has heard — a prompt-layer change that inherits the two-green-live-run cost, per the issue and the doc.


AlaskanTuna:
## Formats Named, and a Correction to My Own Justification

Pushed as `081c110`. 1099 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

### 1. The Option Descriptions Now Name Formats

```
attachment_1..3   Share a file with Roka: PNG, JPEG, GIF, WebP, PDF, MP3, WAV, OGG, FLAC, AAC or AIFF (1 of 3)
                  Another file: PNG, JPEG, GIF, WebP, PDF, MP3, WAV, OGG, FLAC, AAC or AIFF (2 of 3)
attachment_url    Link to a PNG, JPEG, GIF or WebP for Roka to look at
```

92, 82 and 52 characters. The list is extensions rather than MIME types precisely because of the 100-character cap. `attachment_url` names its four formats too, which is better than "an image" — it stays honest about being narrower than the slots beside it *and* tells you what it takes.

**The drift guard matters more than the wording.** `src/discord/__tests__/ask.test.ts` derives the expected labels from `ALLOWED_IMAGE_TYPES` / `ALLOWED_DOCUMENT_TYPES` / `ALLOWED_AUDIO_TYPES` and asserts each is named in every slot, so adding a type to the sets fails the suite until the description is updated. It also asserts the link option does **not** name PDF or any audio format, since naming one it would refuse is the over-promise in the other direction. Pinned by mutation: admitting an unnamed `audio/opus` fails 1, dropping FLAC from the description fails 1, adding PDF to the link option fails 1.

### 2. You Are Right, and I Have Corrected the Reasoning

`geminiMimeType`'s comment claimed sending Discord's name unchanged "is a guess". Your measurement shows it is not — `audio/mpeg` is accepted by `generateContent`, not merely by `countTokens`, with identical counts. That justification was wrong and is now replaced with what you measured: **both labels are accepted, Gemini routes on content rather than the declared type, and the rename is a deliberate preference for the documented name** — which carries a compatibility promise an undocumented alias does not — rather than a workaround for a rejection. Same correction in `docs/trd.md` and in the test comment that said "renamed rather than guessed at either end".

One thing to flag on your report: `docs/trd.md` did **not** actually carry the false claim — it stated the naming facts and that the rename happens, without asserting the alias would fail. Only `attachmentLimits.ts:34` asserted it. I have added your measurement to the TRD anyway, because it is worth recording, but the correction was one file rather than two.

Your caveat is recorded rather than quietly used: `audio/mp4` being accepted on mislabelled MP3 bytes tests the label, not AAC-in-MP4 content. The refusal of `audio/mp4` stands on the Discord-layer reasoning and I have not leaned on your probe for it.

### 3. A Test I Wrote and Then Deleted

I first added a test asserting each description stays within 100 characters. Mutating a description to 120 characters produced **no test failure and no test run at all** — `SlashCommandBuilder.setDescription` throws at construction, so an overlong description makes the module unimportable and the suite fails to collect before any assertion executes.

So that test could only ever pass. It is removed rather than kept, with a comment in both files recording that the builder enforces the cap structurally and harder than a test could. The mutation probe is the only reason I know that — it was the one probe whose result I could not predict, and it turned out the thing I was measuring was already guarded by something better.

### On #131

Not folded in here, per your note. Worth saying plainly that if attachment bytes really are re-sent as history, this PR's `tokensInEst` figure for audio — deliberately 0 — understates by a different mechanism than the one I documented, and the comment explaining *why* it is 0 would need revisiting. That is a reason to measure #131 before #128, not a reason to hold this.

PR #133, opened 2026-08-19
fix(agent): stop re-sending attachment bytes as conversation history every turn

### LINKED ISSUE #131: chore(agent): attachment bytes stay in session history and are re-sent every turn
Found while answering an operator question about whether Roka stores the files people send her. The storage answer is reassuring; the cost consequence is not.

## What Happens To Attachment Bytes

**Nothing is written to disk, and nothing is written to SQLite.** `saveMessage` persists a text `content` column only, and rehydration rebuilds events as `parts: [{ text }]`, so a restart drops attachments entirely. There is no `writeFile`, no temp file, no blob column anywhere on the attachment path.

**But they persist in memory, and they are re-sent.** `@google/adk`'s runner appends the incoming message to the session verbatim — `runner.js:125`, `content: newMessage` — and `newMessage.parts` is exactly where `roka.ts:864` pushed the `inlineData`. Nothing strips it afterwards; `inlineData` appears once in `roka.ts` and that is the push.

So a posted image stays in the live session's event list, and **every subsequent turn in that channel re-sends it to Gemini as conversation history**, until one of:

- it falls out of `WindowedSessionService(config.session.windowSize * 2)` — 40 events, roughly 20 turns;
- the idle TTL destroys the session — `session.ttl`, 500 s;
- the bot restarts.

## Why It Matters

#121 priced an image at a flat ~1089 tokens and concluded three images add 59% to *that turn*. That figure counts the turn carrying the attachment and no others. If an image survives ~20 turns of history, its true amortised cost is a multiple of what was measured — one image posted into an active channel could cost 1089 tokens **per turn** for the rest of the window, not once.

That does not overturn #121's verdict on its own: RPM still binds before TPM, and the worst case there was already far beyond observed usage. But it means `tokensInEst` still understates image turns even after #124 instrumented them, because #124 counts only the attachments arriving *this* turn.

## What To Check First

**Confirm the resend actually happens before designing anything.** The reasoning above is code-read, not observed. `usageMetadata.promptTokenCount` is already on the response object at `roka.ts:495` but is only used in a warn log for empty responses — capturing it would settle this in one turn: post an image, then send a text-only follow-up in the same channel, and see whether the second turn's prompt tokens carry the image.

That measurement is worth having regardless. It is the difference between an estimate and the truth for every token figure in `response_events`.

## Options, If Confirmed

- **Strip `inlineData` from session events once the turn completes**, leaving the text marker. She keeps knowing an image was there; she stops paying for it every turn. Closest to how rehydration already behaves after a restart, so the two paths would finally agree.
- **Leave it and document it**, if the model genuinely needs to refer back to the picture across turns — which is a real conversational capability, not obviously worth discarding.

The choice is a product one, not a technical one, and it should be made deliberately rather than inherited from what ADK happens to do by default.

AlaskanTuna:
Closes #131, using the measurement already on that issue rather than repeating it.

## What Was Happening

ADK's runner appends the incoming message to the session verbatim and nothing in the framework removes it, so `inlineData` parts stayed in the live event list and were **re-sent to Gemini as conversation history on every later turn** — until they aged out of the 40-event window, the 500 s idle TTL fired, or the bot restarted. Measured: ~11.5 MB of heap retained per attachment, accumulating turn over turn.

`WindowedSessionService.stripAttachmentBytes` now replaces those bytes with a text marker — `(an image)`, `(an audio clip)`, `(a document)` — once the turn is over.

## The Three Things That Were Not Obvious

**1. Where to strip.** The runner appends `newMessage` *before* the agent runs, and the request contents are built from session events. Stripping at append time would mean she never sees the attachment at all — a silent break that looks exactly like working code. So the strip runs after the turn.

**2. After every attempt, never between them.** A retry re-sends the same message, so ADK appends it again — one upload can leave several copies in history, and stripping mid-loop would hand the model a marker where the first attempt had the picture. All copies are tracked and stripped, pinned by a test asserting a retried turn strips 2.

**3. Why holding the reference works.** `appendEvent` receives the very object pushed into storage — neither it nor `createEvent` clones. `getSession` *does* deep-clone, so the obvious approach of fetching a session and stripping it would mutate a copy and change nothing. That asymmetry is why the strip lives in the session service rather than at the call site.

It also runs on **failed** turns, since a turn that errored still appended its message.

## The Capability This Removes — Named, Not Inherited

She can no longer refer back to a picture from an earlier turn. She knows one was there; she cannot see what was in it. That is a real loss and worth stating plainly.

The reason to accept it: **rehydration after a restart already rebuilds history as text-only.** So the capability was never reliable — it existed or not depending on how recently the bot had been deployed, and the live path and the restart path disagreed about whether an image from four turns ago was still visible. They now agree. Recorded in `docs/trd.md` rather than left implicit.

## Verification

1065 tests pass (+14), `test:perf` passes, tsc / Biome / Prettier clean.

**The control is the important test.** `sessionAttachmentRetention.contract.test.ts` drives a *real* ADK `Runner` against a capturing model, and asserts first that an attachment **is** re-sent on the next turn when nothing strips it. Without that, the fix would be unfalsifiable — if ADK had never re-sent attachments, the stripped case would pass while proving nothing. This repo now has its own evidence the bug was real, independent of the measurement on #131.

It deliberately avoids `__setTestRunTurnFactory`: that seam replaces the call to `runner.runAsync`, so `appendEvent` never runs beneath it and a retention test written against it observes nothing whether or not retention exists.

Six mutation probes, counts predicted first — all matched:

| --- | --- | --- |
| Strip does nothing | 6 | 6 |
| Only the first copy is stripped | 1 | 1 |
| Every attachment called an image | 2 | 2 |
| `appendEvent` tracks nothing | 6 | 6 |
| `deleteSession` keeps tracking | 1 | 1 |
| **Call site removed** | 2 | 2 |

### The Last Row Is the One Worth Reading

After the first five probes passed I checked whether deleting the **call site** in `generateResponse` broke anything. It broke nothing — all 1063 tests passed with the strip never invoked. Every retention test drove the session service directly, so the fix was fully tested and completely unwired-checked.

Closed by exporting `sessionService` (same convention as `rokaAgent`) and two tests asserting `generateResponse` reaches the strip on a succeeding turn and on a failing one. The probe now fails 2, as predicted.

### One Test Bug Worth Recording

The first version of the retention test used a module-level `withImage` fixture. Since the fix works *by mutating shared references*, the first stripping test mutated the fixture every later test reused, and four tests failed for reasons unrelated to the code. The captured `LlmRequest` objects had the same problem — they hold live references to the parts being stripped, so a kept request appeared never to have carried the attachment. Fixtures are now built per use and the capturing model snapshots each request at generation time.

## Interaction With #130

`docs/multimodal.md` now records that its per-turn token figures hold *because* of this strip. Before it, they understated by up to 20×. That matters most for audio and video, which are billed per second — a five-minute clip riding twenty turns was the worst case in the codebase.

This branch is off `main`, so it does not carry #130's audio types; the marker function already handles `audio/` for when that lands.


AlaskanTuna:
## Reviewed By Hand — Approving The Substance, Two Notes

**The unwired-fix finding is the most valuable thing in this PR.** A fix that is fully tested and never invoked is worse than an untested one, because the green suite actively argues against looking. Probing the *wiring* rather than the logic is what caught it, and that is the third instance of this family now — the `NaN` in #129's guard, my stubbed heap probe, and this. The common shape is that the assertion and the production path had drifted apart while both looked healthy.

**The control test is what makes this reviewable.** Asserting first that `turns[1].inlineMimes` is `['image/png']` *without* the strip means the repo now owns evidence the bug was real, independent of my live measurement. Had ADK never re-sent, the stripped case would have passed while proving nothing.

**Reference-holding vs `getSession`'s deep clone.** Verified, and it is the non-obvious constraint that justifies putting this in the session service rather than at the call site. Worth the comment it got.

### Note 1 — The Strip Is Not In A `finally`

`stripAttachmentBytes` runs after `runTurnWithReliability` returns. That function converts model failures into a `fallbackResult` rather than throwing, so the strip is reached on every ordinary path including failures — which is why the wiring test for a failing turn passes.

But an *unexpected* throw from inside it would skip the strip, leaving `attachmentEvents` holding references to events that may since have aged out of the window. It self-heals: the next turn on that channel strips everything tracked, and `deleteSession` clears the map, which the idle TTL guarantees within 500 s. So the exposure is bounded and small.

Not blocking, and I would not restructure the retry loop for it. Flagging because the bound depends on `runTurnWithReliability` never throwing, which is a property of that function today rather than a guarantee of this one.

### Note 2 — Needs `main`

`DIRTY`. #130 merged since you branched, so `roka.ts` now carries the audio types and `attachmentLimits.ts`'s `geminiMimeType`. Take `main` in — merge, not rebase — and verify by content afterwards rather than trusting a clean auto-merge, the way you did on #129. Both changes touch `downloadAttachment`'s neighbourhood.

`attachmentMarker` already branches on `audio/`, so the marker side is ready; it is the surrounding file that moved.

### On The Design Decision

Accepting the loss of cross-turn image recall is the right call and the reasoning is the one that matters: the capability was never reliable, only a function of how recently we deployed, because rehydration has always rebuilt history as text-only. Making the two paths agree turns an accident into a stated contract.

I am **not merging this yet** — it is a user-visible capability change rather than an invisible fix, so it goes to the operator rather than being shipped on my judgement. Everything else here is ready.

AlaskanTuna:
## Took Main — and This Time There Really Was a Conflict

Merged `origin/main` (not rebased), pushed as `9235550`. 1113 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

**`docs/trd.md` conflicted for real**, unlike the #129 prediction — both #130 and this branch inserted a new section immediately before `### ADK Error Delivery Constraint`. Both sides were purely additive, so both are kept, ordered so what is admitted comes before what becomes of it: **Attachment Types and Their Ceilings**, then **Attachment Bytes Do Not Live in History**.

`src/agent/roka.ts` auto-merged, which is the part worth distrusting rather than trusting. Verified by content:

- #130 survived — `sizeLimitFor(contentType)` at the download limit, `geminiMimeType(contentType)` on the passthrough branch, `MAX_AUDIO_SIZE_BYTES` intact.
- This branch survived — `attachmentMarker`, `attachmentEvents`, `stripAttachmentBytes`, the `appendEvent` override and the call site all present.
- Both test suites collect.

And re-ran the two probes that matter most afterwards, because a merge can unpin tests as easily as it can break code: strip-does-nothing still fails 6, call-site-removed still fails 2. Both as predicted.

Audio now exercises the marker end-to-end rather than in anticipation — `(an audio clip)` was already asserted here, and `audio/ogg` is a real admitted type as of #130.

## The `finally` Nit — Declined, With the Bound Written Down

You are right that the strip is not in a `finally`, and right about why it does not bite today. I have not restructured, for two reasons:

**The exposure is smaller than "skipped".** If an unexpected throw skips the strip, the events stay *tracked*. The next turn in that channel strips them along with its own, and `deleteSession` clears them when the TTL fires. So the cost is **one extra resend**, not a permanent leak — bounded twice over.

**The diff would not be surgical.** The reliability expression spans ~100 lines; wrapping it means re-indenting all of it, so reviewers read a hundred changed lines to cover a path no test can reach.

What I have done instead is write the dependency down at the call site, in your terms rather than mine — that the bound rests on `runTurnWithReliability` converting failures rather than throwing, that this is a property of *that* function and not a guarantee of this one, what it costs if that changes, and that the fix is to revisit here if it ever gains a throwing path. An implicit cross-function dependency is the thing worth removing; documenting it precisely does that, where a comment saying "safe" would not have.

Say the word if you would rather have the `try`/`finally` regardless and I will do it.

## On the Hold

Agreed, and I would not want it merged on peer judgement either. A capability a user could notice going away is not "not breaking", and reading it that way would stretch the operator's authorisation past what they said. The framing you put to them is the honest one: this makes an accident into a stated contract, because rehydration has always rebuilt history text-only.

If the answer is no, the alternative is not "revert" — it is bounding the retention deliberately, e.g. keeping bytes for one or two turns rather than the full window, which gets most of the saving while keeping short-range recall. Worth having in reserve.

Issue #132, opened 2026-08-19
feat(agent): the prompt never tells Roka she can hear or watch anything, so she deflects


AlaskanTuna:
Found while testing #127's intake path live. The plumbing works; the experience does not.

## What Happens Today

A real 6-second MP3 attached to a turn, asked directly to describe it:

> **"Describe exactly what you hear in this clip — is it a tone, music, or speech?"**
>
> `(´・ω・｀) Hmm~? Mio, are you playing mysterious sounds for me at this hour? It's already past midnight, you know~ 🌙`

She acknowledges *something* arrived and describes nothing. That is a deflection, not an answer.

**The model is definitely receiving it.** `countTokens` on the same bytes returns 67 tokens for a 2-second clip — 2 × 32 tok/s plus overhead — so Gemini decoded real audio. This is not a perception failure.

## The Cause

`src/agent/prompts/core.ts` contains **zero** occurrences of "audio", "hear", "sound", "listen", "video" or "watch". It has an `## Images` section:

> - When users share images with you, acknowledge and comment on what you see naturally — react as yourself, not as an AI describing an image

…and no counterpart for anything else. Meanwhile `## Hard Boundaries` says:

> - If asked about something outside your world that no tool can answer, deflect naturally in character

With no instruction that hearing is something she does, a clip looks like exactly that case, and the deflection rule is the one that fires. The reply above is the prompt working as written.

## Why It Matters Now

#127 ships audio intake and #128 ships video. Both will land a capability whose entire user-visible surface is "she says something vague and changes the subject" — indistinguishable from the feature being broken. The silent-drop problem from #100 was hard to diagnose for the same reason: nothing about the failure told anyone what had failed.

`docs/multimodal.md` deliberately scoped this out:

> Intake only. If Roka needs telling *how* to talk about a video she has watched, that is a prompt-layer change and inherits the two-green-live-run cost that #94 and #105 paid — it should not be bundled here.

That call was right, and this issue is not asking to reverse it. It is asking to stop the deferral being invisible: intake without it is a feature nobody can tell is working.

## Scope

A modality section in `CORE_PROMPT`, sibling to `## Images`, covering what she does with a clip she has heard or watched — the same shape as the existing image guidance, not a new behavioural system.

[...]

PR #134, opened 2026-08-19
feat(agent): let Roka watch a short video behind a streaming size guard

### LINKED ISSUE #128: feat(agent): let Roka watch a short video, low resolution only, behind a streaming size guard

AlaskanTuna:
Closes #128 — steps 5 and 6 of `docs/multimodal.md`, the last of the #100 split.

**Based on `fix/strip-inline-data-from-history` (#133), not `main`**, because video is the modality where re-sending attachment bytes as history hurts most: a 60 s clip is ~6,000 tokens, and riding twenty turns would be 120,000 against a 250,000 TPM ceiling from one upload. Merge #133 first and this retargets to `main` cleanly.

## Part 1 — The Streaming Guard

`readWithinLimit` streams the body with a running byte counter and cancels the reader the moment it passes the ceiling. The `content-length` pre-check stays as a cheap early exit; it is what refuses an honestly-declared oversized file before a byte is read.

**There is deliberately no `arrayBuffer()` fallback for a body-less response.** A fallback is a path where the guard does not run, and it is exactly the path a malformed response would take.

That decision had a visible cost: every existing attachment test failed immediately, because the `serve()` stub returned `{ ok, headers, arrayBuffer }` and no body. That is the guard working — the stub was fixed to serve a real chunked `ReadableStream`, which is also what makes the abort observable.

## Part 2 — Video, Low Resolution Only

Ten types admitted, 10 MB cap, and `video/quicktime` admitted alongside `video/mov` for the same reason `audio/mpeg` was: it is the registered type for a `.mov` and what Discord reports, while Gemini's list says `video/mov`. Renamed at the download boundary.

**Media resolution is pinned per request, not on the agent — a deliberate departure from the AC's wording.** The AC says "media resolution is pinned to low explicitly", which reads as agent config. But `mediaResolution` is *request-level* and governs "tokens per input image **or video frame**" — setting it on `rokaAgent` would have silently re-priced and re-rendered every image she can already see, changing a shipped feature to implement a new one. `beforeModelCallback` now sets it only when the request actually carries a video part. The AC's real requirement — "the request asserted in a test rather than the setting assumed" — is met either way, and this way the assertion is meaningful rather than incidental.

## Verification

1153 tests pass (+40), `test:perf` green, tsc / Biome / Prettier clean.

| --- | --- | --- |
| Picker forgets the video formats | 10 | 10 |
| Running counter never trips | 8 | **4** |
| Media resolution pinned unconditionally | 2 | 2 |
| `video/quicktime` not renamed | 1 | 1 |
| Video falls through to the document ceiling | 0 | **0** |

### The Two Rows That Are Not Ticks

**The counter probe came back 4, not 8, and the reason matters.** I predicted every "refuses an oversized X" test would fail. Only four did — and they are precisely the ones whose `content-length` is absent or understated. The honest-header cases never reach the counter at all, because the pre-check short-circuits them first. So the size tests are split across two mechanisms, and I had been attributing all of them to the guard. The four that genuinely exercise it are the four written for it; nothing is unguarded, but the coverage is narrower than the test names suggest at a glance.

**The video ceiling branch is unobservable, predicted and confirmed.** `MAX_VIDEO_SIZE_BYTES` and `MAX_DOCUMENT_SIZE_BYTES` are both 10 MB, so deleting video's branch from `sizeLimitFor` changes no behaviour and no test can fail. The branch is kept because the two ceilings are set by different arguments and a later change to one should not silently move the other — but the test asserting it is weak by construction, and both the test and the constant now say so rather than implying a guard that is not there.

### A Gap in My Own Drift Guard

The `/ask` description test from #130 derived its expected labels from `ALLOWED_IMAGE_TYPES`, `ALLOWED_DOCUMENT_TYPES` and `ALLOWED_AUDIO_TYPES` — an enumeration that must itself be maintained. Adding video left ten types admitted and unnamed in the picker **with the whole suite green**, which is the exact failure the test was written to prevent.

Fixed at the root: `attachments.ts` now declares `ALLOWED_MEDIA_TYPES` as the union, `isSupportedMedia` tests membership in it, and the drift guard reads it. A fifth set added to the union is covered automatically; one forgotten is not admitted at all, which is a loud failure rather than a silent one.

## Not In Scope

Duration limits, per the issue. And **#132** — the prompt never mentions hearing or watching, so she will deflect if asked about a clip. This ships intake, not the ability to tell it is working; the AC should not be read as delivering a user-visible feature until #132 lands.


AlaskanTuna:
## An Acceptance Criterion I Had Missed, Found by Reading #135

Pushed `e6aa5ea`. 1157 tests pass (+4), `test:perf` green, tsc / Biome / Prettier clean.

Reading #135 to check whether it moved this PR's caps — it does not, the caps are unchanged and prefixing changes what happens *above* them — its opening line did something more useful. It says oversized media is **refused silently**. #128's acceptance criteria say *"An oversized video is refused in character, never dropped silently."* I had not implemented that.

What I shipped refused the bytes correctly and told nobody. `unsupportedCount` in both handlers is derived from the **type** check only, so an oversized file — a type she takes, bytes she cannot — passed that check, died at the download, and vanished. She would answer the text and never mention the file: the exact silent-drop shape as #100, #114 and #118.

`generateResponse` now returns `droppedAttachments`, and both handlers nudge on `unsupportedCount > 0 || droppedAttachments > 0`.

### Written as an OR Rather Than a Sum, Deliberately

The first version was `unsupportedCount + droppedAttachments`. Every existing handler test failed at once, because their `generateResponse` doubles predate the field: `0 + undefined` is `NaN`, `NaN > 0` is `false`, and the nudge disappears entirely — including for the unsupported-type case that already worked. That is the #129 `NaN` shape again, caught by tests this time rather than shipped.

The doubles are fixed regardless, in all four suites. But the comparison is now an OR, so an absent field degrades to *the previous behaviour* rather than to silence. Failing back to "nudges on type only" is recoverable; failing to "never nudges" is the bug this whole PR is about.

### Two Test-Isolation Problems Fixed On the Way

- My regex for completing the doubles also matched `toolsUsed: []` inside two `expect.objectContaining` blocks asserting `recordResponseEvent`, which does not receive this field. Four metrics tests failed; both insertions removed.
- `describe('unsupported attachments on the mention path')` had **no setup of its own** and was inheriting whatever the previous block last left on the mocks. My new test set a return value and silently changed the meaning of the test after it. It now has the same `beforeEach` as its siblings, which fixes the pre-existing fragility as well as mine.

### Probes

| --- | --- | --- |
| Dropped count always zero | 2 | **1** |
| `/ask` ignores dropped downloads | 1 | 1 |
| Mention path ignores dropped downloads | 1 | 1 |
| Every download counted as dropped | 1 | 1 |

The first was my arithmetic error, not a coverage gap: forcing the count to `0` cannot fail a test that expects `0`. That made me check whether the control was load-bearing at all, so I mutated in the opposite direction — counting every download as dropped — and it fails only the control. The pair pins the count from both sides.

## On #135

Nothing here needs to change for it, and I have not tried to anticipate it. The caps stay 8 MB and 10 MB; prefixing replaces the *refusal* above them, not the number. Two things from this PR that #135 inherits cleanly:

- `reservationFor` already clamps a stated size to the type ceiling, which is #135's last acceptance criterion. It was added in #129 for a different reason — the download refuses on `Content-Length` before buffering — and a `Range` fetch makes it true for a second, better reason: the capped size really is what arrives. Worth asserting there rather than assuming, as the issue says.
- `droppedAttachments` is the seam for #135's "say so in character". A prefixed file is not dropped, so it will want its own signal — but the plumbing from the agent layer back to a user-visible line now exists and is tested, which it was not this morning.

The transcode-versus-prefix framing on #135 is right, and the arithmetic that whole-file ingestion of 200 MB audio is *impossible* rather than merely expensive is the part that settles it.

AlaskanTuna:
## Untangled Against the Squashed Main — Mergeable

Pushed `0c4c123`. `mergeable=MERGEABLE`, base `main`, and the diff is 15 files that are all this PR's.

The tangle was that this branch carried #133's **pre-squash** commits while `main` carries the squashed version, so the same changes existed on both sides from a merge base of #130. One real conflict, in `src/agent/roka.ts` — `attachmentMarker`'s video branch and `requestCarriesVideo`, where this branch is a strict superset of what #133 landed. Resolved by keeping the superset.

Then verified by content rather than trusting the resolution, since "both sides had it" is exactly the case where a resolution can quietly drop one:

| From | Symbol | Present |
| --- | --- | --- |
| #133 | `stripAttachmentBytes`, `attachmentEvents` | yes |
| #130 | `geminiMimeType`, `MAX_AUDIO_SIZE_BYTES` | yes |
| #134 | `readWithinLimit`, `requestCarriesVideo`, `MEDIA_RESOLUTION_LOW`, `MAX_VIDEO_SIZE_BYTES`, `droppedAttachments` | yes |

`attachmentMarker` has all four branches. And the check that actually settles the duplication question: **the diff against `main` re-adds none of #133's symbols** — `grep -c` over added lines for `stripAttachmentBytes|attachmentEvents|WindowedSessionService` returns 0. So the merge kept one copy, not two.

Re-ran the four probes afterwards, because a merge that resolves in favour of one side can unpin the tests guarding the other: counter 4, media resolution 2, dropped count 1, strip 6. All unchanged from before the merge.

1157 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

## On the Check Status

`gh pr checks` now says it in as many words:

```
CodeRabbit  pass  Review skipped: manual review required for this OSS repository
```

A **pass** that means no review happened. That is the behaviour worth keeping in view whenever this repo's green tick is used as evidence — the tick and the review are unrelated here, and the tick is the more visible of the two. This PR's only review is the human one.

PR #138, opened 2026-08-19
feat(agent): take the opening of an oversized file instead of dropping it

### LINKED ISSUE #135: feat(agent): ingest a prefix of oversized media instead of dropping it silently
Operator question that prompted this: *"what if someone uploads a 200 MB audio/video (the person has Nitro) into /ask… I don't expect Roka to ingest the entire audio/video if it's too large, unless you have a method to do it economically."*

There is one, it is cheaper than expected, and it needs no transcoding.

## Today: Oversized Media Is Refused Silently

`downloadAttachment` refuses on `content-length` before buffering, so a 200 MB upload costs nothing — verified, `arrayBuffer` is never called and the heap does not move. But the attachment then simply **does not exist** as far as the turn is concerned. She answers the text and never mentions the file. Same silent-drop shape as #100 and #118: nothing about the failure tells anyone what failed.

## The Method: Take A Byte Prefix

Measured against a real 4 min 36 s MP3 (2.77 MB, 84 kbps). A raw `head -c` cut — no re-encode, no container repair:

| Bytes taken | Tokens | Result |
| --- | --- | --- |
| 2.77 MB (whole) | 8,845 | full summary |
| **1.00 MB (prefix)** | **3,175** | *"Anthropic claims to have found a 'bizarre global workspace' hidden deep inside Claude's brain."* |

Correct, quoted, and proportional — 1 MB of an 84 kbps stream is about 100 s, and 100 s × 32 tok/s ≈ 3,175. **MP3 is frame-based, so any prefix is valid audio.**

Video too, when the container allows it. The operator's 28 MB Senren\*Banka OP is fragmented MP4 — `moov` sits at byte 36, followed by `moof`/`mdat` pairs — so a 6 MB prefix was accepted at 2,757 tokens against 11,431 for the whole file.

**The caveat that decides the design:** this only works when `moov` precedes the media data. A plain MP4 straight off a phone often carries `moov` at the end, and a prefix of one is undecodable. The atom layout is readable from the first few bytes, so it is checkable cheaply — but the fallback when it fails must be a refusal, not a corrupt upload.

## Why It Is Cheap

Paired with an HTTP `Range` request, **the excess is never downloaded at all**. A 200 MB file costs one ranged GET of 8 MB — no transcode, no `ffmpeg` in the image, no CPU on the Pi, and the existing byte budget and per-type caps keep working unchanged because what arrives really is 8 MB.

Contrast the transcode option costed on #128: that needs `ffmpeg` in the container and CPU time inside the turn. Prefixing needs neither. Transcoding preserves the *whole* clip at lower quality; prefixing preserves *part* of it at full quality. For a 200 MB podcast the second is plainly the better trade.

## Duration Is The Real Ceiling Anyway

TPM is 250,000 input tokens/minute (measured, #125), and production averages ~5,543 text tokens per turn. Even ignoring memory:

| | Full ingest of 200 MB | Fits in one turn's TPM |
| --- | --- | --- |
| Audio (32 tok/s) | ~5.3 h ⇒ ~611,000 tokens — **impossible** | ~2.1 h at 100% of TPM |
| Video (~91 tok/s) | — | ~45 min at 100% of TPM |

A single turn should never eat a whole minute's budget, so the practical figure is far lower — at a 20% share, roughly 26 min of audio or 9 min of video. **Whole-file ingestion of very large media is not merely expensive, it is arithmetically impossible.** A bounded prefix is the only shape that works.

## Proposed Behaviour

1. Attachment within cap → unchanged.
2. Attachment over cap → `Range`-fetch the first `MAX_*_SIZE_BYTES`, verify the container tolerates a prefix, and send that.
3. **Say so in character.** "I only listened to the first few minutes" is a good answer; silence is not. This is the part that distinguishes it from today's behaviour, and it should not be dropped as polish.
4. Container check fails (`moov` at the end) → refuse, and say that too.

## Acceptance Criteria

- A file over the cap yields a prefix of exactly the cap, not a refusal, when the container permits.
- The reply names the truncation. A test asserts the turn carries the marker rather than only that bytes were sent.
- A non-prefixable MP4 is refused audibly, not silently, and never sent truncated.
- `Range` is verified against Discord's CDN specifically — the whole saving rests on the excess never crossing the wire.
- The byte budget reserves what will actually arrive (the capped size), not the stated size. `reservationFor` already clamps to the ceiling, so this should need no change — assert it rather than assume it.

AlaskanTuna:
Closes #135, built on the measurements already on that issue rather than repeating them.

A file past its ceiling is now `Range`-fetched down to exactly that ceiling and sent as its opening, and she says she only got through the beginning. Today it vanishes and she answers as though nothing were attached.

## Where It Refuses, and Why That Matters More Than Where It Prefixes

**A wrong prefix is not an error.** It is a request that succeeds and an answer about nothing — no exception, no failed turn, nothing in the logs. So the policy refuses unless the prefix is provably readable:

| Type | Policy | Reason |
| --- | --- | --- |
| `audio/mpeg`, `audio/mp3` | prefix | Self-describing frames, no trailing index — any prefix is valid audio |
| `mp4`, `mov`, `quicktime`, `3gpp` | prefix **iff** `moov` precedes the media data | A property of the file, not the format |
| OGG, WebM, FLAC, AAC, WAV, AVI | refuse | Plausibly prefixable; **none measured** |
| Images, PDF | refuse | A truncated PNG or PDF is corrupt, not partial |

`isobmffAllowsPrefix` walks the top-level box list of the bytes **already fetched**, so the container check costs no extra request. It handles 64-bit box sizes, a size-0 box that runs to EOF, and non-box bytes — all of which return `false`, because "cannot prove" has to mean refuse.

Leaving OGG and WebM out is deliberate. They are very likely fine, and that is exactly the reasoning this repo has been punished for all week. They belong in the list with a measurement behind them, not an argument.

## Three Details That Carry the Saving

**A file refused on its stated size is never requested.** A 200 MB PDF costs zero transfer, asserted by a test that the fetch stub was called zero times.

**`Range` is an optimisation, not a correctness requirement.** A server may ignore it and answer 200 with the whole body. The first version treated that overflow as a failure — so a Range-ignoring server turned the saving into a refusal, silently. `readWithinLimit` now takes an overflow mode: `refuse` for an ordinary download, `truncate` for a deliberate prefix, where reaching the ceiling is the plan. Both paths are tested, including that the cut lands on **exactly** the ceiling rather than wherever the chunk boundary fell.

**She says so, separately.** `truncatedAttachments` is its own count, not folded into `droppedAttachments` — a turn can carry one file she could not open and another she could only start, and a test asserts both lines appear.

## Verification

1196 tests pass (+39), `test:perf` green, tsc / Biome / Prettier clean.

| --- | --- | --- |
| Never prefix anything | 5 | 6 |
| Skip the container check | 1 | 1 |
| Send no `Range` header | 1 | 1 |
| `/ask` silent about truncation | 2 | 2 |
| Treat every type as prefixable | 3 | 10 |

Both misses are in the *more coverage than counted* direction — the `it.each` blocks over refused types multiply a single mutation into many failures. Worth stating rather than rounding to "all passed".

## The Acceptance Criterion I Could Not Meet

> `Range` is verified against Discord's CDN specifically — the whole saving rests on the excess never crossing the wire.

**Not done, and I cannot do it from here.** It needs a live `cdn.discordapp.com` attachment URL, and those are signed and expiring — I have none to hand and cannot fabricate one. What is verified is that we send `Range: bytes=0-<limit-1>`, and that the read stops at the ceiling whether or not the server honours it.

So the *correctness* of this PR does not depend on Discord supporting `Range`; only the bandwidth saving does. If the CDN ignored it entirely, oversized files would still be prefixed correctly — we would just have read the bytes and thrown them away rather than never requesting them.

**If someone can paste a real Discord attachment URL, this is a one-request check** and I will run it and report the status code and `content-range`. Until then it should be treated as designed-for but unconfirmed, not as met.


AlaskanTuna:
## The AC Is Answered, and the Answer Inverts the Design's Assumption

Measured here rather than taken on report, since it decides which code path actually runs:

```
GET https://cdn.discordapp.com/embed/avatars/0.png   Range: bytes=0-99
  → 200 (not 206), accept-ranges: bytes, content-length: 1268, cf-cache-status: HIT
  → 1268 bytes delivered. Asked for 100, got the whole file.
```

**Discord's CDN advertises `accept-ranges` and then ignores it.** So the branch I described as defensive — a server answering 200 with the whole body — is not a fallback. It is the only path that runs in production. The first version of this PR treated that overflow as a failure, which would have turned *every* oversized file into a silent refusal on the one path that exists. That test earned its keep, and it is renamed to say so rather than reading as an edge case.

### And the Correctness Argument Was Understated

I wrote that if `Range` were ignored "we'd just have read the bytes and discarded them rather than never asking". That undersells it, and the difference is measurable, so I measured it:

```
cancel at 1 MB of a 50 MB body → 1.00 MB in  178 ms
full download of the same body → 50.00 MB in 2750 ms
```

Cancelling the reader **stops the transfer**. We do not read 200 MB and keep 8 — we read 8 and hang up. So the saving is structural in the cancel, with `Range` as a bonus that Discord declines to grant. `Range` stays because it costs nothing and a 206 would be strictly better.

Adopting the suggested rewording, which is provable where the original was not:

> `Range` is sent, and the read stops at the ceiling whether or not it is honoured — verified against `cdn.discordapp.com`, which advertises `accept-ranges` and ignores it.

**One caveat kept honest:** that avatar is a 1,268-byte cached object, and Cloudflare commonly ignores `Range` on small or cached bodies. A multi-megabyte attachment on a cache miss may behave differently. A 50 MB Cloudflare object also answered 200 here, but from a different origin, so it does not settle Discord's behaviour on a real upload. Read this as **`Range` cannot be relied on**, not as *`Range` never works* — and it does not matter either way, because the cancel is what bounds the transfer.

Code comment, `docs/trd.md` and the test name all corrected to say the cancel is the guarantee.

## A Red Commit, and a Flake It Exposed

`0314422` was pushed with a failing test. My command chain ran the commit regardless of the test result rather than gating on it — my mistake, caught immediately, and it never reached `main`.

The failure was worth having. `strips history even when the turn failed` — my test from #133 — timed out at 5 s. It passes 3/3 in isolation: it drives a **real** retry ladder with real backoff, ~2 s idle and over 5 s on a loaded machine. A test that depends on wall-clock backoff is a flake waiting for a slow CI box, so `liveMaxRetries` is now taken out for that one turn. The retry count was never what it asserted.

1196 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

## On Documents Not Deflecting

Useful boundary for #132: reading text was never the problem, which is why #119 shipped without anyone noticing. Worth keeping in view here too — the partial-ingest line this PR adds is written for media she *perceives*, and a truncated document does not reach it, because documents are refused rather than prefixed. Nothing to change; just not a gap hiding behind the same words.

AlaskanTuna:
## Reviewed By Hand — One Real Gap, Then Ship

CodeRabbit says `Review skipped: manual review required for this OSS repository`, so this is the only review it gets.

The module is the most careful thing either of us has written today. Every entry in `STREAMABLE_TYPES` traces to a token ratio, the two counter-intuitive survivors are documented as *why* rather than *what*, and the ffmpeg caveat is stated rather than buried. `'streamable'` over `'frames'` is the right rename — it was a small lie for WAV and WebM, and precision in a policy name is worth more than tidiness.

### The Gap: `moov` Found Is Not `moov` Intact

`isobmffAllowsPrefix` returns `true` the moment it reads the box **type**, before establishing that the box **body** fits inside the prefix:

```js
if (type === 'moov') return true
```

A file whose `moov` starts early but extends past the cut passes the check while the prefix carries a truncated index. That lands in exactly the failure this module exists to prevent, and which its own docblock names: *"no error anywhere — the request succeeds and the answer is about nothing."*

Narrow, but worth closing because of which way it fails. It needs a `moov` larger than the prefix cap, so 8 MB+ of index — a very long or very complex file. Rare, and **currently those files are silently dropped, which is not a wrong answer.** This change would turn a silent drop into a confident answer about nothing, so on this specific path it would be a regression rather than an improvement.

Suggested fix, in the shape the rest of the walk already uses:

```js
if (type === 'moov') return offset + size <= prefix.length
```

with `size` resolved before the check rather than after. A test cutting a fragmented MP4 mid-`moov` pins it — and it should assert `false`, since "cannot prove it" is the module's stated default.

### Two Smaller Notes, Neither Blocking

**The 64-bit comment is slightly off.** It says *"reading the low half is enough to keep walking"*, but `readBigUInt64BE` reads the full 64-bit value and `Number()` converts it, with `Number.isSafeInteger` catching anything implausible. The code is right; the comment describes a different implementation. Worth a word since the guard below it is what actually does the work.

**`declared === 0` returns `false`.** Correct — a to-end-of-file box has nothing after it to find — and I mention it only because it is the one branch where refusing looks like an oversight rather than a decision. The comment covers it.

### On The Flake Audit Coming Back Empty

Running per-test durations rather than guessing which tests looked risky is the right method, and reporting "one member, already fixed, not filing an issue" is the right outcome. Filing work that is already done would have been worse than not filing.

The `trialPacing` figure you flagged has a history you did not have: **that exact test flaked on me this morning** in a full run, passed in isolation and on two clean re-runs, and I attributed it to `vi.stubEnv` racing across files. Your mechanism — a cold `await import()` after `resetModules`, 3,588 ms against a 5,000 ms default, first of four cycles — fits every observation and mine did not. Yours is the correct diagnosis and mine was a guess that happened to be unfalsifiable. Please do the one-line file timeout as its own change; it is a real intermittent failure with a known cause, not speculative hardening.

PR #140, opened 2026-08-19
fix(agent): tell the model when an attachment could not be retrieved

### LINKED ISSUE #137: fix(agent): when an attachment fails to download she invents one rather than saying so
Found while measuring #132. **Pre-existing** — baselined against the prompt without #132's changes and it reproduces identically, so it is neither caused nor worsened by that work.

## What Happens

When an attachment fails to download, `imageParts` is empty and the turn proceeds as if no file was ever attached. The user's text still says "watch this", so the model has a request to describe something and nothing to describe. It invents one, confidently.

Two runs, identical setup — a `video/mp4` attachment whose download returns 404, message *"watch this and tell me what happens in it"*:

> "The video is a rapid-fire **19-minute recap** of Stephen King's **IT**, tracing the story from the children's terrifying encounters with the shape-shifting clown **Pennywise** in Derry in the late 1980s through to the adult reunion **twenty-seven years later**…"

> "**MIO: Memories in Orbit** follows a tiny robot navigating the haunting corridors of an abandoned vessel, repairing a massive mechanical hand, activating a hidden wheel, and unlocking the flowing steps ability…"

Different fabrications, same failure. Neither hedges.

Notably the **audio** case handles it correctly — *"did you forget to actually attach or play it for me?"* — so this is not uniform across modalities, and video is the reliable reproducer.

## Why It Is Not Fixed By Telling Her Not To

#132 adds *"If you genuinely cannot make anything out, say so plainly. Never invent what was in it."* It does not help, and cannot: from the model's position there is no failed attachment to notice. The request looks like an ordinary question about a video, and nothing in the context contradicts that. **This is a missing-input problem wearing a prompt-problem costume.**

## The Fix Is Plumbing, Not Wording

#134 already computes `droppedAttachments` and surfaces it to the *user* as a nudge. The model is not told. Passing the same signal into the turn's context — a line stating that a file was attached and could not be read — gives the model something true to say, and makes the existing "never invent" instruction actionable instead of decorative.

That seam is new as of #134 and is already tested from agent layer to user-visible line; this extends it one hop further, to the model.

## Why It Matters More Now Than It Did

- Audio (#127) and video (#128) both shipped today, so there is far more that can fail to arrive than there was this morning.
- A download failure is exactly when a user is *most* likely to ask what she saw, because from their side the file uploaded fine.
- #135 will make partial ingestion routine, widening the class of "the model has less than the user thinks it has".

A confident fabrication is the worst failure shape available here: silent drops (#100, #114, #118, #134) at least fail quietly. This one asserts something untrue in her own voice and gives the user no way to tell.

## Acceptance Criteria

- A turn whose attachment failed to download carries that fact into the model's context, not only into the Discord reply.
- With the attachment missing, she says so — and a test asserts the reply does **not** describe content, rather than asserting it contains particular words.
- The successful path is unchanged and carries no extra context line — assert that directly, since the cheap way to pass this is to always inject it.
- Inherits the prompt-change cost if the wording lands in `CORE_PROMPT`: two consecutive green live tool-trigger runs (#94, #105).

AlaskanTuna:
Closes #137. The model is now told that a file failed to arrive, so it has something true to say instead of inventing a film.

## One File Changed, and Not a Prompt One

The diff touches `src/agent/roka.ts` only. `droppedAttachments` already existed from #134 and already reached the *user* as a nudge; this carries the same signal one hop further, into the turn's parts:

```
[1 file(s) were shared with this message but could not be retrieved.]
```

**Deliberately a statement of fact, not an instruction.** What to do about it is already covered by "never invent what was in it"; the reason that line was decorative is that from the model's position there was nothing to notice. This supplies the missing input rather than adding a rule.

That distinction is also what keeps the change cheap: no wording lands on the prompt path, so it does not buy the two-green-live-run cost — and a line that only materialises when a download has already failed could not affect a tool-trigger gate anyway, since no fixture has a failing download.

## A Duplicated Branch, Removed Rather Than Tested

The notice initially went into both arms of `buildNewMessage` — the ordinary path and the safety-ladder path that drops images. A mutation deleting it from the **`dropImages` arm alone broke nothing**: a second copy that no test could have caught going wrong.

Rather than write a test that drives safety escalation to rung 3 to cover it, the branch is gone. The parts list is built once, with `dropImages` deciding only whether the images go in. The safety ladder drops the pictures; it never had a reason to drop the explanation of why they are missing.

## Verification

1207 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

| --- | --- | --- |
| Never add the notice | 2 | 2 |
| Always add the notice | 2 | 3 |
| Drop it from the `dropImages` arm | 0 | **0** — the finding above |

The "always add" probe matters as much as the first: the cheap way to pass a test that looks for the line is to inject it unconditionally, so the quiet paths — attachment arrived, no attachment at all — are asserted directly.

## The Half I Have Not Verified

The acceptance criteria ask that *she says so*, asserted as the reply not describing content. **That is a behavioural claim about the model and it is not tested here** — a unit test scripts the reply, so it can prove the line reaches the model and nothing about what the model does with it.

I have the check written and have deliberately not run it: it is two `generateContent` calls, a control without the notice and the same turn with it. The reason for holding is not quota — it is that a live tool-trigger gate is running on the same key right now, free tier is 15 RPM, and two stray calls landing in the wrong minute could 429 a half-finished gate that costs far more than this does.

It runs the moment that key is clear, and the result goes on this PR either way. Until then this should be read as **plumbing proven, behaviour unproven** — which is the honest state, and given the day's pattern of things that are fully tested and do nothing, not a state I want to leave unstated.


AlaskanTuna:
## Verified Live — and the Mechanism in #137 Is Not What the Issue Says

Ran the behavioural check. It took three attempts to get a **valid** one, and the first two are the interesting part.

### Two Controls That Proved Nothing

**Attempt 1** — a synthetic three-line system prompt. The control answered *correctly*: "you didn't attach or link any video for me to see!" No fabrication. So the experiment showed nothing: the notice cannot be credited with fixing a bug the control never exhibited.

**Attempt 2** — the **real** assembled prompt, all 14,564 characters. The control still declined gracefully: "Since I can't actually see moving pictures through a screen like that…"

Two failed reproductions meant a missing variable, not a fixed bug. The clue was in the issue itself: the fabrications were **real things**. "MIO: Memories in Orbit" is a real game; a 19-minute IT recap is a real video. Those are not invented from nothing — they look like search results.

### Attempt 3, With Tools Declared

```
CONTROL 1  tool calls: search_web { query: "Mio watch this and tell me what happens in it video" }
CONTROL 2  tool calls: search_web { query: "watch this and tell me what happens in it" }
CONTROL 3  tool calls: search_web { query: "watch this and tell me what happens in it" }
CONTROL 4  tool calls: search_web { query: "Mio watch this and tell me what happens in it" }

NOTICE 1-4 tool calls: none
           "Mou~, Mio! You can't just tease me with a file you didn't actually send…"
```

**4 of 4 versus 0 of 4.**

So #137's fabrication is not free hallucination. The model **searches the web for the user's phrasing and reports what it finds as though it had watched the file**. The issue says "it invents one, confidently" — accurate about the output, wrong about the cause, and the cause is what makes the fix work.

The notice does not merely give her something true to say. **It stops the search from happening at all**, because the request no longer looks like a question about a video that exists. That is a better fix than the issue anticipated, and it is also why cutting "do not describe or invent their contents" cost nothing: the fabrication was never a disobedience problem.

### What This Changes Here

One test added, pinning that the notice sits **before** the request in the parts list. Measurement makes that ordering load-bearing rather than cosmetic — the line has to be in front of the request it contradicts.

1208 tests pass, `test:perf` green, tsc / Biome / Prettier clean.

### Cost and Caveats

Twelve `generateContent` calls on the development key, run after the RPM arithmetic was checked against the live gate rather than guessed at. Four samples per condition at temperature 0.9 — enough to say *consistently*, not enough to say *always*, and the harness declares a minimal `search_web` rather than the full `rokaTools` set, so it demonstrates the mechanism rather than reproducing the production tool surface exactly.

The behavioural half of the acceptance criteria is now met. Worth noting it was met by an experiment whose first two attempts were invalid — and both looked like passes.

PR #142, opened 2026-08-19
feat(agent): refuse attachments that cost more than one turn may spend

### LINKED ISSUE #136: feat(agent): admit attachments by measured token cost, not by size — a 17 KB PDF can cost 28,000 tokens
Operator question: *"should we implement rate limit for this audio/video attachment ingestion? Same goes to general images and PDFs."*

Measured the answer per modality. It is **yes, but not a blanket one** — and the modality most exposed is the one already shipped.

## Token Cost Is Not Bounded By The Byte Cap

| Modality | Cost driver | Measured | Byte cap bounds it? |
| --- | --- | --- | --- |
| Image | flat per image | **1,089** | irrelevant — flat regardless of size |
| Audio | duration | **32 / second** | weakly — 8 MB is 8.5 min at 128 kbps but ~33 min at 32 kbps |
| Video | duration | **~91 / second** | weakly — same bitrate dependence |
| **PDF** | **pages** | **560 / page** | **not at all** |

The PDF row is the problem. A **17 KB** 50-page PDF costs **28,001 tokens** — three orders of magnitude more token cost than its bytes suggest. `MAX_DOCUMENT_SIZE_BYTES` bounds nothing that matters here, and PDFs shipped in #119.

## What That Permits Today

Worst case, one user at `rpm: 15` with `MAX_ATTACHMENTS = 3`, against the measured 250,000 TPM ceiling (#125):

| | Per turn | Per minute | % of TPM |
| --- | --- | --- | --- |
| Images | 8,810 | 132,150 | **53%** |
| Video, 10 MB @ 2 Mbps | 16,463 | 246,945 | **99%** |
| Audio, 8 MB @ 128 kbps | 54,503 | 817,545 | **327%** |
| PDF, 200 pages | 341,543 | 5,123,145 | **2049%** |

And a single request can breach the ceiling on its own:

```
3 x 200-page PDF = 341,543 tokens in ONE request  -- exceeds 250,000 TPM
3 x 500-page PDF = 845,543 tokens in ONE request
```

That is not a rate-limit problem, it is an **admission** problem: no rate limiter helps when one request is already over budget. The turn fails with `429 RESOURCE_EXHAUSTED`, the reliability layer retries it with backoff, and the retries fail identically — while consuming the minute's TPM for **every other channel**, not just the sender's.

`byteBudget.ts` does not help either: it bounds memory, and a 17 KB PDF is invisible to it.

**Images are fine and need nothing.** #121 established RPM binds before TPM with ~2x headroom, and the 53% row is that finding restated. A blanket per-attachment limit would tax the one modality that is already safe.

## Proposal: Admission By Measured Cost, Not By Count

**`countTokens` is free and exact.** Verified twice today: it does not draw on the generate quota — it answered normally on a key whose 500 RPD generate budget was exhausted — and it returns the true figure for any modality without parsing containers, counting PDF pages, or probing durations.

So, for turns carrying attachments only:

1. Call `countTokens` on the assembled request.
2. Refuse or truncate above a configured per-turn ceiling, **in character** — "that document is far too long for me to read in one sitting" is a good answer; a `429` retry storm is not.
3. Track spend against a rolling per-user token budget so one person cannot monopolise TPM with individually-legal files.

Costs one extra round trip on attachment turns only — a few hundred ms, no CPU, no new dependency. Text turns are untouched.

### Why Not The Alternatives

- **Per-user attachment count** — blunt. Punishes images (1,089) identically to a 200-page PDF (112,000), a 100x difference.
- **Tighter byte caps** — does not address PDFs at all, and would refuse legitimate audio to solve a problem audio does not have.
- **Page/duration parsing** — needs a PDF parser and `ffprobe` in the image, and gets a worse answer than `countTokens` gives for free.

## Acceptance Criteria

- A turn whose measured cost exceeds the per-turn ceiling is refused **audibly and in character**, never silently dropped and never sent to fail on `429`.
- The ceiling and the per-user window are tunable in `config.yml` with inline comments, bounded in `NUMERIC_BOUNDS`.
- `countTokens` failure is non-fatal: fall back to admitting the turn rather than blocking on the probe. A guard that fails closed here turns a Gemini hiccup into a total attachment outage.
- Text-only turns take no extra round trip — assert that directly, since the cheapest way to pass this issue is to probe every turn.
- A test pins the 560/page and 32/s figures as the *reason* the ceiling exists, so a future change to `MAX_DOCUMENT_SIZE_BYTES` does not read as if it bounds cost.

## Related

- #135 (prefix ingestion) handles *oversized* media; this handles media that is legal in bytes but expensive in tokens. A 17 KB PDF passes every byte check there is.
- #121 measured the image case and correctly concluded no action. This does not reopen that.

AlaskanTuna:
Closes #136.

## The Problem, In One Number

**A 17 KB PDF costs 28,001 tokens.** Measured on a synthetic 50-page document: Gemini bills PDFs at ~560 tokens per page, and page count has no relationship to file size. `MAX_DOCUMENT_SIZE_BYTES` bounds memory and nothing else.

Cost per modality, all measured against `gemini-3.5-flash-lite`:

| Modality | Driver | Cost | Bounded by the byte cap? |
| --- | --- | --- | --- |
| Image | flat | 1,089 each | irrelevant — flat at any resolution |
| Audio | duration | 32/second | weakly — 8 MB is 8.5 min at 128 kbps, ~33 min at 32 kbps |
| Video | duration | ~91/second | weakly, same bitrate dependence |
| **PDF** | **pages** | **560/page** | **not at all** |

Worst case at `rpm: 15` with `MAX_ATTACHMENTS = 3`, against the measured 250,000 TPM ceiling (#125): images reach 53%, audio 327%, a 200-page PDF **2,049%**. And a single turn of three 200-page PDFs is **341,543 tokens — over the ceiling in one request**, where no rate limiter helps: it fails on 429, the reliability ladder retries into the same wall, and it spends the minute's TPM for **every other channel**, not just the sender's.

## The Fix: Price It, Then Decide

`countTokens` is the only thing that prices all four modalities without parsing containers, counting pages or probing durations — and it **does not draw on the generate quota**, verified twice today against a key whose 500 RPD was exhausted and which still answered here normally.

- Turns carrying a non-image attachment are priced before sending; over `gemini.maxAttachmentTokens` the attachments are refused and she says so in character, via a new response pool distinct from "unsupported" and "partial".
- **Images are never priced.** A flat 1,089 each caps a full turn at ~3,267, which cannot reach any sane ceiling, so asking would spend a round trip to learn something already known. Text-only turns are untouched.
- **The probe fails open.** A guard that blocked the turn when it could not answer would convert a Gemini hiccup into a total attachment outage — worse than the single over-budget request it exists to prevent.

Default ceiling **50,000**: a fifth of TPM, which admits three maximum-length audio clips and refuses a 200-page PDF.

## Notes For Review

**The floor is derived, not written out.** ~~`min` is `3267`~~ — the peer pointed out that `src/agent/attachmentLimits.ts` has **zero imports**, which is exactly the shape needed. `GEMINI_IMAGE_TOKENS` moved there, `imageProcessor.ts` re-exports it for its existing callers, and `config.ts` now reads `min: MAX_ATTACHMENTS * GEMINI_IMAGE_TOKENS` directly. The cycle (`imageProcessor → logger → config`) is not closed because nothing on the new path imports anything. The `config.test.ts` arithmetic test stays, but it now checks a derivation rather than guarding a literal against it.

**Pricing costs a second upload.** `countTokens` sends the same base64 the message will send, so a priced attachment crosses the wire twice — at the Pi's measured ~2.5 MB/s that is ~4s each way for a 10 MB PDF, inside the 20s timeout with room but real latency on the largest admitted files. It is the concrete reason the image skip earns its keep rather than being a micro-optimisation. Recorded in `docs/trd.md` and in `attachmentCost.ts` so it is not discovered later on a latency graph.

**The image skip is model-specific, not merely model-derived.** It rests on 1,089 being *this* model's flat price, and `gemini.model` is configurable. A model that priced images by resolution would not make the skip stale — it would make it wrong, waving through the one case it exists to bound. Noted in the code; the derived floor above at least keeps the ceiling moving with the constant.

**The probe is a `fetch`, which broke two existing tests.** #138's prefix tests count fetch calls, and a real `countTokens` adds one per non-image turn. `attachmentCost.js` is now mocked in `roka.test.ts`. Flagging it because it is the same shape that bit me twice today: a stub and the production path drifting apart.

## Verification

`tsc` clean · Biome clean · Prettier clean · **1221/1221** (from 1202).

Three mutation probes, failure count predicted before each run, all three matched:

| --- | --- | --- |
| Guard removed — never refuse | 1 | 1 |
| Probe every turn, images included | 1 | 1 |
| Fail closed on probe error instead of open | 1 | 1 |

No live gate needed: no prompt file is touched, so tool-triggering is unaffected.

## Merged With Main, And One Interaction Nobody Flagged

Both sides added code at the same point in `roka.ts` — this branch's admission block and #140's `failedAttachmentNotice`. Kept both, admission first, and verified by content rather than by a clean `git status`.

Resolving it surfaced a real interaction between the two changes. **A cost refusal removes the attachment, which re-creates exactly the condition #140 had just fixed**: the file is gone, the request still says "what happens in this video", and `CORE_PROMPT` tells her to quietly `search_web` a fact she is unsure of — measured at 4 of 4 fabrications before #140 and 0 of 4 after. So the notice now covers both reasons an attachment can be absent, worded apart because they are not the same fact:

- `[N file(s) were shared with this message but could not be retrieved.]`
- `[N file(s) were shared with this message but are too long to read in one turn.]`

A test pins the second one, since it is the arm that did not exist before this merge.


AlaskanTuna:
## Reviewed — Approving the Design. One Number Independently Confirmed, One Correction That Is Mine, Two Things To Fix.

### I Re-Measured Your Load-Bearing Number, and It Holds

The whole ceiling rests on 560 tokens/page, so I priced it myself with `countTokens` on PDFs I generated with known page counts rather than taking the figure on report:

```
blank,  1 page              561 tokens
blank, 10 pages            5601 tokens   → 560.0/page marginal
text-heavy, ~300 words/pp  5601 tokens   → text adds exactly 0
```

**560.0 per page, and page content is irrelevant.** Your number is right, and the second row is the part neither of us had: a dense page and a blank page cost the same, so cost tracks page *count* and nothing else. That makes your case stronger than the PR states — page count is not merely unbounded by bytes, it is the *only* variable.

### And It Shows a Document I Wrote Is Wrong

`docs/multimodal.md` says **258 tokens per page** and *"text already embedded in the PDF is extracted and not charged as tokens. So a text-layer PDF is close to free."* Measured: 560, and a text-layer PDF is **not** close to free — it costs exactly what a blank one does. My error from the #100 research, taken from Google's docs rather than measured, and it has been sitting in the file that every one of tonight's caps was derived from. Fixing it separately so it does not ride in on your PR.

### Fix Before Merge

**1. It will conflict.** Your diff carries the pre-#140 two-arm `buildNewMessage`; `main` now has the single-list form. Take `main` and expect the conflict there — and verify by content afterwards, since both sides touch the same expression.

**2. The literal 3267 is avoidable, and the tool already exists.** `src/agent/attachmentLimits.ts` has **zero imports** — I created it as a pure leaf in #130 for exactly this problem, a constant needed by a module that could not import the heavy one. Move `GEMINI_IMAGE_TOKENS` there and `config.ts` can import it directly with no cycle. `MAX_ATTACHMENTS` can follow the same way if you want the whole product derived rather than half of it, with `attachments.ts` re-exporting so no call site changes — it is a per-turn ceiling both layers use, not a Discord-specific one.

### Your Skip Is Right, and I Checked It Properly

You asked me to test whether it is right rather than merely clever. It is, and the reason is `needsMeasuring` returning true if **any** part is non-image: a mixed turn is measured, and `countTokens` receives the whole parts array, so the images are priced too. The skip only ever applies where every part is flat-rate.

Two things to know rather than change:

- **It rests on 1,089 being this model's number.** `gemini.model` is configurable; if it changes, image-only turns silently stop being bounded, and nothing anywhere would notice — the skip is the one path that can never be measured.
- **`MAX_ATTACHMENTS × 1089 = 3267` is not `>` a ceiling set to 3267**, so the boundary is safe at the floor rather than one short of it. Worth a test if you have not got one.

### The Cost Nobody Has Stated

`measureAttachmentTokens` sends the full base64 payload, so pricing a 10 MB PDF **uploads it twice** — once to ask and once to send. At the Pi's measured 2.5 MB/s upstream that is roughly 4s + 4s inside a 20s `gemini.timeout`, and the doubling lands hardest on exactly the largest attachments. It is inherent to the approach and I would not change it; `countTokens` cannot price bytes it has not been given. But it belongs in the PR and in `docs/trd.md` as a known cost rather than something the next person discovers from a latency graph.

### Fails Open — Agreed, and For the Reason You Give

A guard that blocks when it cannot answer converts a hiccup into an outage. It is also the right asymmetry here specifically: the failure it prevents is one over-budget request, and the failure it would cause is every attachment turn failing.

Nice piece of work. The measurement in the PR body — 17 KB costing 28,001 tokens — is the kind of number that makes an argument unnecessary.

AlaskanTuna:
## Re-Reviewed at `fba0544`. Your Question Answered With the Mechanism, Gates Verified Independently, One Gap Found.

### The `dropImages` Question — Yes, and Here Is Why, Not Just That

Traced rather than reasoned about. The line numbers are the answer:

```
1035  let dropImages = false
1076  let refusedAttachments = 0          ← set here, before the loop
1106  const failedAttachmentNotice = [...] ← computed here, before the loop
1122  const buildNewMessage = ...          ← closure, re-read per attempt
1171  dropImages = true                    ← set here, inside escalateSafety, during the loop
```

`failedAttachmentNotice` is a `const` fixed **before** the reliability loop starts; `buildNewMessage` re-reads `dropImages` on every attempt but the notice array is already frozen. So a safety escalation to rung 3 cannot remove it. **The notice survives.**

And there is a second reason it cannot break, which is stronger than the first because it does not depend on ordering: a refusal sets `imageParts.length = 0`, so by the time `dropImages` could matter, `dropImages ? [] : imageParts` evaluates to `[]` either way. The two paths do not interact at all after a refusal — one of them is already a no-op.

### Gates Verified Independently

Checked out `fba0544` and ran them here rather than reading the numbers: **tsc clean, Biome clean, 1221/1221**. Matches.

### The Refusal Re-Creating #137 Is the Best Catch of the Exchange

That is a failure I introduced the conditions for and did not think to look for in your path — #140's notice fired on `droppedAttachments` alone, so your refusal would have walked straight into the 4/4 fabrication. Wording the two arms apart is right: *never arrived* and *arrived and cost too much* are different facts, and collapsing them would have made the second one a lie.

### The Gap: the Count Is Not Pinned, and I Think It Is Also Wrong

Probed your branch rather than reading it. Removing the refused arm of the notice fails **1** — pinned. But:

```
refusedAttachments = imageParts.length   →   refusedAttachments = 1
```

**fails nothing.** 712/712 still pass. The count is free to be any number.

That matters more than an untested line usually would, because I do not think `imageParts.length` is the right number. **The refusal is all-or-nothing**: a turn carrying one cheap image and one 500-page PDF refuses both, and then reports *"2 file(s) … are too long to read in one turn"* — of which one was 1,089 tokens and perfectly affordable. The count is right about how many files did not reach her and wrong about why, in her own voice, to the person who sent them.

Two ways out, and I do not think this blocks:

- **Keep all-or-nothing and fix the sentence.** The set is what exceeded the ceiling, so say that rather than attributing it to each file. Cheapest, and honest.
- **Refuse only what is expensive.** Costs a probe per attachment, which given your double-upload finding is exactly the wrong direction.

I would take the first, plus a test on the count so the next person cannot quietly change it. With `MAX_ATTACHMENTS = 3` the common turn carries one file and never notices — which is precisely why it would sit there.

### The Rest

The literal is gone properly — `min: MAX_ATTACHMENTS * GEMINI_IMAGE_TOKENS`, derived rather than asserted, and `attachmentLimits.ts` stays a zero-import leaf so nothing new can close a cycle through it. The double-upload cost is recorded in the shape I would have written it, and *"a model pricing images by resolution would not make the skip stale, it would make it wrong"* is a sharper statement of the risk than mine was.

Approving on the substance. The count is worth one commit before it goes.

PR #165, opened 2026-08-20
fix(agent): allow for the video audio countTokens does not price

### LINKED ISSUE #153: fix(agent): the attachment cost estimate ignores a video's audio track, which the model still hears
## `countTokens` Does Not Count A Video's Audio Track, But The Model Hears It

Found while answering an operator question about whether Roka perceives sound in video. Both halves were measured, not reasoned.

### She Hears It

A 20-second video of **pure black frames** with speech as its only audio, so nothing in the picture could reveal the content. Asked to transcribe anything spoken:

> “Yesterday, Anthropic published a new paper claiming it found a bizarre global workspace, hidden deep inside the brain of Claude…”

Correct, and it could only have come from the soundtrack. **A video part carries both frames and audio to the model.**

### The Estimate Does Not Count It

Same clip, with and without its audio stream, priced by `countTokens` against `gemini-3.5-flash-lite`:

| Input | Tokens |
| --- | --- |
| 20 s clip, video + audio | **2,133** |
| 20 s clip, video only (`ffmpeg -an`) | **2,133** |
| 20 s of the same audio as an MP3 | 643 (32.1/sec) |

**Identical.** The audio track adds exactly zero to the estimate, while the same audio priced on its own is 32 tokens a second.

### Why It Matters

`src/agent/attachmentCost.ts` prices a turn with `countTokens` and refuses above `gemini.maxAttachmentTokens`. For video that estimate is **missing ~32 tokens per second of whatever the audio would have cost** — around a quarter of a low-resolution video's measured ~107/sec.

So the admission check is **too permissive on video, never too strict**. A clip priced at 49,000 tokens could bill nearer 63,000 if the audio is charged. At `MAX_VIDEO_SIZE_BYTES` of 10 MB this is bounded and not urgent, but it is a known gap between what the guard measures and what the request does — the same family as everything on #141.

### What Is Not Established

Whether Google *bills* the audio. `countTokens` is already known to diverge from billing for time-based media — it over-reports video by ~65% in earlier measurements — so its silence here is not proof the audio is free. Two possibilities, both consistent with the evidence:

1. Audio in video is genuinely not billed, and the estimate is right.
2. Audio is billed and `countTokens` omits it, so the estimate is low.

The conservative reading is (2), because the failure direction is the dangerous one.

### Suggested Fix

Add a duration-derived audio allowance to the video estimate — or simply bound video harder than the generic ceiling, given that video is the one modality where the estimate is known to be incomplete in the permissive direction. Either is small. Deciding between them wants a billing-side measurement rather than another `countTokens`.

*Measured 2026-08-20 with `ffmpeg` to build matched clips and `countTokens` to price them; the behavioural test used one `generateContent` call.*

AlaskanTuna:
Closes #153.

## Reproduced Independently Before Fixing

The issue's `countTokens` half re-measured with clips I built rather than its numbers — black frames plus a **sine tone**, not speech, so the content shares nothing with the original:

| Clip | Bytes | Tokens |
| --- | --- | --- |
| 20 s video + audio | 190,831 | **2070** |
| 20 s video only (`ffmpeg -an`) | 10,276 | **2070** |
| that audio alone, as MP3 | 321,557 | 643 (32.05/s) |

**18× the bytes, zero extra tokens.** The audio rate reproduces the issue's to two decimals (32.05 vs 32.1); the video rate differs (103.5/s vs 106.65/s) because the clips differ, which is expected — the *identity* is the finding, and it holds on both pairs.

The behavioural half is taken from the issue's measurement rather than re-run, deliberately: it costs a `generateContent`, and both keys are budgeted for the #132 gate.

## The Fix

`measureAttachmentTokens` adds `VIDEO_AUDIO_ALLOWANCE` (0.31, from 32.05 ÷ 103.5) when a video part is present. `MAX_ATTACHMENTS` is 1, so the whole measured total is that video's.

**What is not established, and is said plainly in the code and the TRD: whether Google *bills* the audio.** `countTokens` already diverges from billing for time-based media — it over-reports video by ~65% in earlier measurements — so its silence here is not evidence the audio is free. Two readings fit the evidence, and this takes the conservative one because only one direction fails badly:

- **under-estimate** → admits a turn that then 429s, retries into the same wall, and spends the minute's budget for every other channel
- **over-estimate** → refuses one long video early

Applied to silent video too. `countTokens` cannot distinguish one, and finding out means parsing a container per format — mp4 and webm are different problems. Over-charging a silent clip by 31% errs toward refusing, which costs a message rather than the minute.

## What Would Retire This Constant

One `generateContent` per clip, reading `usageMetadata.promptTokenCount` — the *billed* figure — would replace the allowance with a fact and could remove it entirely. Two calls, decisive. Not run here because the gate has the quota; flagged in the code so it does not become folklore.

## Tests

`adds an allowance for the soundtrack countTokens does not price` — 2,070 → 2,712.

`reports %s exactly as the API priced it` over audio, PDF and image — **the control**. An allowance applied to everything would pass the first test while being wrong, so the types `countTokens` prices completely must come back untouched.

## Gates

`npm run typecheck` 0, `tsc --noEmit` 0, Biome 0, Prettier 0, 1289/1289, perf 2/2.

| --- | --- | --- |
| allowance → 0 | 1 | 1 |
| apply the allowance to every type | 4 | 4 |

Restored, md5 matched.


AlaskanTuna:
Added the wire, per review: `rests on a turn carrying exactly one attachment`.

The allowance inflates the **whole** measured total, which is only the video's cost because a turn carries exactly one attachment — and nothing in either file linked those two statements. At `MAX_ATTACHMENTS` of 3, one video beside two images would apply 31% to the images too. Safe, but no longer what the comment claims, and it would read as intentional long after it stopped being true.

Written out rather than derived, per #160: raising the constant now fails here instead of silently over-charging.

Probe: `MAX_ATTACHMENTS` 1 → 3, predicted 1 failure in `attachmentCost.test.ts`, got 1. Restored, md5 matched.

Gates: typecheck 0, tsc 0, Biome 0, Prettier 0, 1290/1290, perf 2/2 — each read as an exit code rather than inferred from quiet output.

PR #170, opened 2026-08-20
fix(utils): reserve a turn's calls so rpm bounds requests, not turns

### LINKED ISSUE #167: fix(utils): the rate limiter meters turns while the quota meters requests, so rpm 15 can issue up to 60 calls a minute
## `rateLimit.rpm` Bounds Turns. Gemini Counts Requests.

`tryConsume()` is called **once per turn**, at three entry points:

```
src/discord/events/messageCreate.ts:446
src/discord/events/interactionCreate.ts:113
src/discord/events/toolCommands.ts:25
```

A turn is not one request. `src/agent/roka.ts:1232` passes `runConfig: { maxLlmCalls: config.gemini.maxLlmCalls }`, which is **4**. A turn that calls a tool issues an initial model call, then another after the tool result, chaining up to that ceiling.

So `rpm: 15` admits 15 **turns** a minute, and those turns may issue up to **60 requests** a minute against a 15 RPM quota.

This is the third instance of one shape: a guard metering a different unit than the quota it defends. #136 metered bytes when the quota was tokens. #149 metered capacity when the quota was a rate. This one meters turns when the quota is requests — and it is the guard we most believed in, because it is named after the quota.

## Two Measurements, Deliberately Kept Apart

**Production** — read-only on the Pi, `response_events`, 302 turns spanning 2026-07-22 to 2026-08-20 (measured by @rokabot-53):

| Tools fired | Turns |
| --- | --- |
| none | 263 (223 `[]` + 40 `NULL`) |
| 1 | 38 (recall_user 19, search_web 17, get_current_time 2) |
| 2 | 1 (search_anime + search_web) |
| retries | 45 total — 270 turns at 0, 22 at 1, 7 at 2, 3 at 3 |

Tool-fire rate **12.9%**. Mean **~1.13 calls/turn** before retries, ~1.3–1.4 with them.

**The live tool-trigger gate** — 18 minutes per run, ADK `Sending out request` lines bucketed by wall-clock minute:

| Run | Mean calls/min | Peak | Minutes over 15 |
| --- | --- | --- | --- |
| gate E | 9.2 | 19 | 4 |
| gate F | 9.7 | 17 | 5 |

**These are not interchangeable and the gate figure must not be used as production's sustained rate.** The gate corpus is a tool-trigger benchmark — selected for tool firing is its entire purpose, so its ~1.9 calls/turn describes the benchmark, not the bot. Reusing it as a production rate is the same error as choosing a measurement window to suit the answer, one level up. The production mean is 1.13.

Both are recorded here so neither walks into a later decision alone.

## What Breaks Is The Peak, Not The Mean

The sustained figure is comfortable. `maxLlmCalls: 4` is a configured ceiling and 3.8 calls/turn was observed in the gate, so at 15 turns/min a burst of tool-firing turns reaches ~30 requests/min against a 15 RPM quota — 2× — without the 1.9 needing to be true.

The floor is sized against an average; the cap is enforced against a peak.

## It Feeds Itself

Exceeding RPM produces a 429. `geminiReliability` classifies it `transient_http`, the retry ladder retries, and each retry is another 1–4 requests. **The guard's response to overshoot is more overshoot.** Observed directly in the gate, where the over-cap minutes were contiguous and identical in both runs.

It also compounds with #150: an RPM 429 and a daily-quota 429 are indistinguishable without reading `quotaId`, so the symptom is latency nobody attributes to a rate limiter that reports itself as never having been hit.

## Reachability

**Dormant at current traffic.** Production averages ~10.4 turns a *day*; nothing is failing now. This needs load — a busy channel, or a raised `rpm` — before it bites. Same profile as the RPD-reset half of #150: a real property waiting on traffic.

## Options

- **A — lower `rpm` to `rpm / maxLlmCalls`** (~3 turns/min). Correct, and gives up most of the throughput permanently.
- **B — consume per model call.** Exact, and matches the quota's unit. `roka.ts` already has a `beforeModelCallback` firing before each call, so the seam exists. Rejected: a refusal mid-turn produces a *partial* turn, which is a user-visible failure introduced to fix an invisible one.
- **C — reserve `maxLlmCalls` up front, release the unused.** Mirrors the `byteBudget` reserve/release idiom already in the codebase, adds no seam, and hard-bounds the peak at 15 requests. The release is what separates it from A: A prices every turn at 4 forever, C holds 4 only for the turn's duration. At a measured 1.13 mean and a 12.9% tool rate the sustained ceiling lands near 13 turns/min, so it costs almost nothing in practice.

**C**, on both our readings.

## The Test That Would Have Caught This, And Is Nobody's Today

`rateLimiter.test.ts` asserts admissions against `rpm`, which is exactly what `RateLimiter` promises — a module's own tests cannot catch its contract being the wrong contract for its caller. The missing check lives at the **call site**: requests-to-Gemini per minute, not admissions per minute. Worth adding alongside whichever fix lands, or the next unit mismatch is invisible again.

## A Query Trap For Whoever Reads `response_events` Next

`tools_used` stores both `[]` and `NULL` for "no tool". A naive `tools_used != ''` counts `[]` as a fire and inflates the rate ~6×.

---

*Unit mismatch found from source by @rokabot-29; gate per-minute figures and the production `response_events` measurement by @rokabot-53.*

AlaskanTuna:
Closes #167. Option **C** from the issue, on both our readings.

## What Changed

`tryConsume()` admitted a turn on **one** slot while ADK was free to issue up to `runConfig.maxLlmCalls` = 4 requests for it. Both handlers that reach the model now reserve the ceiling and hand back what the turn did not use, released in the same `finally` as the byte reservation and the concurrency flag.

Calls are counted where they happen — a `modelCallsForRequest` ALS incremented in `beforeModelCallback`, following the existing `toolCallsForRequest` idiom — and returned as `GenerateResult.modelCalls`. Nothing downstream could count them: ADK issues them internally.

## Three Things The Work Surfaced

**The sustained ceiling is `rpm - maxLlmCalls + 1`, not `rpm`.** A turn is admitted only when a *whole* reservation fits, so single-call turns settle at **12/min at `rpm: 15`**, not 15. I predicted 12 in a test, got 9 at `rpm: 12`, and the gap is that standing headroom — a fifth of the budget. That is the number to quote, and it is a real cost of bounding the peak rather than the "almost nothing" we both estimated.

**`rateLimit.rpm` needs a floor of `gemini.maxLlmCalls`.** Below it *no turn can ever be admitted* — the bot answers nothing while reporting itself rate-limited. Found when a harness fixture at `rpm: 2` went silent the moment reservations landed. Now a `NUMERIC_BOUNDS` floor, so it is rejected at startup. Same mutual-satisfiability class as #152.

**A turn that cannot report what it spent keeps its whole reservation.** With an incomplete caller the arithmetic degraded to `NaN` and released nothing *by accident* — right behaviour, wrong reason, and the #137 shape exactly. Now an explicit `Number.isFinite` branch: holding slots costs a minute, handing back spent slots costs the quota.

## Test Doubles Replaced Rather Than Completed

106 tests failed at once on a method the hand-rolled `RateLimiter` doubles had never heard of — and `as unknown as RateLimiter` meant `npm run typecheck` could not see it coming. That is the `vi.fn()`-is-`any` caveat from #151 in another guise.

Where a double only meant "don't interfere", it is now **the real class** with generous limits: it has no I/O, and a hand-maintained mirror of another module is what drifted. Doubles survive only where a test asserts on the interaction (`concurrency.test.ts`, `messageCreate.test.ts`), and those are completed and retargeted from `tryConsume` to `canAdmitCalls`.

Harness fixtures at `rpm: 1` / `rpm: 2` meant "exactly N turns' worth". One turn now costs `maxLlmCalls`, so the meaning is unchanged and only the arithmetic moved. The `rpm: 1, rpd: 0` refusal fixtures are untouched.

## Probes — Two Missed, And Both Misses Improved The Tests

| --- | --- | --- |
| never release | 3 | 5 |
| reserve 1 instead of the ceiling — *first* | 4 | **1** |
| reserve 1 instead of the ceiling — *after* | 2 | 2 |
| `canAdmitCalls` always true — *first* | 2 | **0** |
| `canAdmitCalls` always true — *after* | 2 | 1 |

**Reserve-1 broke only one test, and the reason is the point of the feature.** Net of release, reserving 4 and giving back 3 leaves the *same one slot* as reserving 1 — so every after-the-fact assertion passes either way. What differs is the window **while the turn is in flight**, which is the peak C exists to bound and which nothing was watching. Added `holds the whole ceiling while the turn is still running`, and the probe then predicted 2 and got 2.

**The peek broke nothing**, because `concurrency.test.ts` drives that path through a double — the real `canAdmitCalls` could have returned `true` unconditionally unnoticed. Added real coverage at both levels. It now fails 1, not 2, and that is correct rather than a gap: without the peek the reservation refuses the same turn a moment later, so the outcome is identical and only the message and a Discord round trip differ. The peek is an optimisation, not the guard, and the PR says so.

Restored after each, `md5sum` matched.

## Deliberately Not Fixed

`toolCommands.ts` (`/anime`, `/remind`) calls `tryConsume()` and reaches Gemini **zero** times — verified. It is borrowing this limiter as a generic abuse guard, which is a different purpose that deserves its own decision. Left alone and noted in the TRD.

## Gates

`npm run typecheck` 0, `tsc --noEmit` 0, Biome 0, Prettier 0, **1304/1304**, perf 2/2 — each read as an exit code.


AlaskanTuna:
Both review points fixed on this branch — no need to push to it.

**1. The contradiction is gone.** `docs/trd.md:149` reads **requests** again, and the RPM-Budget Accounting note no longer says the fields are documented in turns "tracked in #167" — #167 is what this closes. It now records that `rpm` is closed and points at the new section.

Worth naming the shape rather than just fixing it: #168 described the world as it was, this PR describes the world as it will be, and both were correct when written. A doc row and the code that contradicts it can each be right at the moment of writing and wrong together an hour later — which is an argument for the row citing the mechanism rather than restating the unit, and it is why line 149 now points "see below" instead of standing alone.

**2. The `rpd` asymmetry is named where it was incurred.** You were right that this was the one decision the TRD explained nothing about, and that `rpd` saying *turns* being newly accurate makes it look deliberate to a reader with no way to tell.

It is now stated twice — once in the new section, once under RPM-Budget Accounting — as a **half-closed gap rather than a choice**: `reserveCalls` increments `dailyCount` once per turn and the release never decrements it, so within one object `admitted` counts requests and `dailyCount` counts turns. `rpd: 500` admits 500 turns, up to 2,000 requests against a 500-request daily quota. Not fixed here, with the reason given: converting it needs release semantics for a counter that has none.

Gates after both: typecheck 0, tsc 0, Biome 0, Prettier 0, **1306/1306**, perf 2/2. Latest main merged in, no conflicts.

